PR #9, opened 2015-12-06
Check expire date of existing certs

tralafiti:


germeier:
Thanks, I was just thinking about implementing this myself.
IMHO the time to renew before expiry should be a config option!

tralafiti:
You are totally right. PR now includes a config option.

lukas2511:
merged, thanks

PR #10, opened 2015-12-06
don't overwrite certificate files

germeier:
In a worst case scenario the new certificate is broken and we are left
without a working certificate (or need to restore one from our backup).

This way we only need to change the symlink to the known working cert


lukas2511:
i merged this in a slightly modified version in b7439a83d6060a1c2c4aee6510c174699f511dbf

PR #13, opened 2015-12-06
show expire date when we don't need to renew a certifcate

germeier:


germeier:
agreed, changed that. Thanks!

PR #14, opened 2015-12-06
Add sed as a parameter variable in order to be able to specify the sed version that must be used.

M-M-M-M:
Add sed as a parameter variable in order to be able to specify the sed version that must be used.

The goal is to be able to specify another sed program as the one to be used must be the GNU sed (-r parameter is used).

For instance default sed program on OS X is not supporting this parameter and launching lets encrypt.sh is displaying following error:

sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]

GNU sed can be installed from macport and its name is gsed. 

So a simple SED=gsed in config.sh file is resolving this issue.

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -14,3 +14,7 @@
 
 # create new private key for each csr (yes|no)
 #PRIVATE_KEY_RENEW=no
+
+# GNU sed program to use - use gsed on some plateforms
+#SED=sed
+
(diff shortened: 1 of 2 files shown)


lukas2511:
I removed the '-r' option from sed as it wasn't needed anyway, as far as i can tell it should be compatible with BSD sed now.

PR #17, opened 2015-12-06
Openssl.cnf parameter

M-M-M-M:
add openssl.cnf as a parameter. 

Default path is not always /etc/ssl/openssl.cnf

It could be some of following:
* OPENSSLCNF=/System/Library/OpenSSL/openssl.cnf
* OPENSSLCNF=/etc/pki/tls/openssl.cnf
* OPENSSLCNF=/usr/lib/ssl/openssl.cnf
* OPENSSLCNF=/usr/local/ssl/openssl.cnf

Goal is to avoid error like:
```

Signing domainwww.example.org (www.example.org)...
cat: /etc/ssl/openssl.cnf: No such file or directory
unable to find 'distinguished_name' in config
problems making Certificate Request
140735281533008:error:0E06D06C:configuration file routines:NCONF_get_string:no value:conf_lib.c:324:group=req name=distinguished_name
```

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -14,3 +14,7 @@
 
 # create new private key for each csr (yes|no)
 #PRIVATE_KEY_RENEW=no
+
+# GNU sed program to use - use gsed on some plateforms
+#SED=sed
+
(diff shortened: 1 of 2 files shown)


lukas2511:
added this in 474f33d2ca79a5852f5a1aa96f0b962ebe267949 with autodetection of default config location

Issue #19, opened 2015-12-07
Certificate revocation


lukas2511:
Revocation of certificates would be a really nice feature to have.

germeier:
I agree!

The protocol spec for "Certificate Revocation" looks straight forward. Anyone who is eager to implement this feature? Otherwise I would give it a try...

lukas2511:
@germeier If you want you can give it a try, shouldn't be too hard.

lukas2511:
implemented with #21 in 063d28a6dcc755d339b05cb55880347a8c6c7b4a

PR #21, opened 2015-12-07
implement revoke

germeier:
as requested by issue #19 

request for comments


lukas2511:
seems to be working, merged in 063d28a6dcc755d339b05cb55880347a8c6c7b4a, thanks.

i think we may need a better way to handle script arguments, and also implement something like "--help", but for now this is good enough

PR #23, opened 2015-12-07
help text and initial code for command parameter

germeier:
an initial try for command parameters and help text

RFC

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -266,10 +266,55 @@ if [[ ! -e "${WELLKNOWN}" ]]; then
   mkdir -p "${WELLKNOWN}"
 fi
 
+
+renew_certs() {
+  # Generate certificates for all domains found in domains.txt. Check if existing certificate are about to expire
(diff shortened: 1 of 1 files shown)


lukas2511:
yea something like that, structure looks fine to me, but i'd prefer the usage of getopts instead of parsing commandline manually

lukas2511:
would suggest following parameters:

### Modes:

```text
-c (cron operation: default)

-s [domain] (force-sign specific domain)

-r [path/to/cert.pem] (revoke certificate)
```

### Parameters:

```text
-f [path/to/config.sh] (use specific config file)

-k [path/to/privatekey.pem] (use specific private key for given operation, especially useful for certificate revocation)
```

germeier:
OK, I will update this pull request tonight.

What does the "-c" mode should do?

lukas2511:
@germeier 
I'd design the '-c'/cron mode to be basically what the normal operation is right now, but only renewing not signing new certificates.

germeier:
I intend to change this pull request to getopt and after we are all happy with the new structure we can add new features using command line options. I already have a few ideas (e.g. using the software without the need for config files)

germeier:
continues on pull request #35

PR #22, opened 2015-12-07
Hook post creation

M-M-M-M:
Add a hook after creation of certificate in order to be able to clean challenge on remote servers and to upload created certs.

PR #24, opened 2015-12-08
add HOOK_POST_CREATION

M-M-M-M:
Add a hook after creation of certificate in order to be able to clean challenge on remote servers and to upload created certs.

To replace pull request #22 with a cleaner commits from last master.

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -17,3 +17,10 @@
 
 # create new private key for each csr (yes|no)
 #PRIVATE_KEY_RENEW=no
+
+# program called after creating the certs arguments: path/to/token token
+# path/to/fullchain path/to/privkey path/to/cert ; can be used to e.g. clean 
+# the challenge or to upload certs if this script doesn't run on the 
+# webserver
+#HOOK_POST_CREATION=
+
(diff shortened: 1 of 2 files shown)


hoedlmoser:
@M-M-M-M I'm not sure what your purpose for this post script is, but in my case I would distribute the certs and keys to all my VMs. so, better to have only certs directory and domain name (because filenames are well known) to be able to create kind of directory structure also on the destination host?

M-M-M-M:
@hoedlmoser  I am using this post script in order to build a chain pem file with intermediate certificates in order to provide it to Haproxy, to deploy this file to all servers by scp, then to send a reload configuration to all Haproxy servers in order to take in account new certificates. Goal is to automate the process on renewal as much as possible.  Croning letsencrypt.sh and that's all. 

Haproxy and Apache need intermediate certificates to avoid warning message on SSL check.

M-M-M-M:
@hoedlmoser And as I am using the HOOK_CHALLENGE to copy by scp the .acme-challenges files I would like to delete it from remote server once certificates have been created.

germeier:
FYI: Your code does not call the hook if there is a problem generating the new cert!

M-M-M-M:
@germeier You are right. I have create pull request #25 to take in account your comment.

I have added the hook before the exit and added a status parameter to the hook program about the valid/invalid challenge result.

silvio:
please close #22 and this one in favour for #25

PR #25, opened 2015-12-08
add HOOK_POST_CREATION

M-M-M-M:
Add a hook after creation of certificate in order to be able to clean challenge on remote servers and to upload created certs.

To replace pull request #24 

Add status parameter and trigger it when challenge failed to answer to @germeier

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -17,3 +17,10 @@
 
 # create new private key for each csr (yes|no)
 #PRIVATE_KEY_RENEW=no
+
+# program called after creating the certs arguments: status path/to/token token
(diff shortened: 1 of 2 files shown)


germeier:
Better, but
on second thought I believe it to be better to only have one hook which can be called with the state from where it is called e.g. "pre", "post", "inthemiddle", ...

Also the challenge may fail due to something that leads to "exit 1" in the _request() function. I suggest to add there also a "post hook" call before exiting to allow a clean up.

M-M-M-M:
@germeier  add call to the _request() in  pull request #28
and a variation with a single hook in pull request #29

PR #27, opened 2015-12-08
Post Challenge & New Cert Hooks

tralafiti:
Another - hopefully cleaner - suggestions for #25.

RFC: Renamed HOOK_CHALLENGE to be in common style. Not sure yet if it is worth breaking compatibility with existing users.


tralafiti:
Closed in favor of #29

PR #35, opened 2015-12-08
help text and initial code for command parameter -- second try 

germeier:
Sorry, the patch is ugly but I had to move some code around.
Now only the handling of the command line is outside of functions

RFC


lukas2511:
reworked this a tiny bit and merged it in 81882a64c2a18df97327980101a547322b70c604.

i think it needs a bit more work, especially on detecting wrong parameters, but it seems to be working for now.

PR #36, opened 2015-12-09
Hook for deploying certs via ssh

digint:
First of all, thanks for this wonderful script!

I implemented a hook for deploying the certificates remotely via ssh, in two parts:

- **ssh_hook.sh**:  the HOOK to define in config.sh.
- **ssh_filter_letsencrypt.sh**: SSH ForceCommand to define in sshd_config (or command="..." in authorized_keys) on the remote webserver.

If you like this patch, I will gladly write some documentation for it (something like the documentation of the ssh script I wrote for btrbk: https://github.com/digint/btrbk#setting-up-ssh).

Maybe it would be good to introduce some "contrib" folder for scripts like these, let me know what you think!

### DIFF
--- a/ssh_filter_letsencrypt.sh
+++ b/ssh_filter_letsencrypt.sh
@@ -1,0 +1,122 @@
+#!/bin/bash
+
+#
+# Use on webserver, in conjunction with ssh_hook.sh, either as
+# "ForceCommand" in "sshd_config", or in "authorized_keys".
+#
+# Example line in "authorized_keys" file:
+#
+#   from="192.168.0.42",command="/home/letsencrypt/ssh_filter_letsencrypt.sh --log",no-agent-forwarding,no-port-forwarding,no-pty,no-user-rc,no-X11-forwarding <pubkey>...
+#
+
+declare -A challenge_dir key_file cert_file fullchain_file
+
+### config section ###
+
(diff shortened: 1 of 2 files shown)


lukas2511:
Mh.. i don't really want to fill the repo with hooks, especially since the script is not yet in a stable form and hook handling may change at any time, requiring to update all example hooks...

Maybe later we can add an "example_hooks" directory, but i'd keep it to 1 script-file per hook.

digint:
> Mh.. i don't really want to fill the repo with hooks, especially since the script is not yet in a stable form and hook handling may change at any time, requiring to update all example hooks...

I understand completely. I'll come back to you when things are more stable, and until then people who are interested in my hook scripts will probably find it via my fork.

> Maybe later we can add an "example_hooks" directory, but i'd keep it to 1 script-file per hook.

Well the second script is here to further lock down the access to the webserver (which is a bit paranoid, but usually that's a good idea), which is done by defining a *ForceCommand* in sshd_config.
A much simpler solution would be to simply scp the challenge and certs, but this would require "full" ssh access, and paranoid people don't like this.

lukas2511:
Closing this for now. If you want to feel free to create a page in the wiki of this repository for your hooks.

PR #39, opened 2015-12-12
remove --sign in favor of two options "--force" and "--domain"

germeier:
I really dislike the current "--sign (-s) domain.tld" command for several reasons:
1. I want to be able to sign a certificate for testing purposes without having to add it to domains.txt. Even one with multiple SAN names!
2. Even when specifying a domain on the command line I want to be informed that the certificate in questing is still within my definied useful time frame (e.g. longer valid than RENEW_DAYS) before really forcing a renew
3. In the long run the letsencrypt.sh script should be able to work properly without any config files only using default values and command line options.

Therefor I suggest the following patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_script:
 script:
   - ./letsencrypt.sh --help
   - mv config.sh tmp_config.sh
-  - ./letsencrypt.sh --sign "${TMP_URL}" -f tmp_config.sh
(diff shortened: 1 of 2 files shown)


lukas2511:
Can you please squash your commits and fix the travis script? Otherwise it's okay to merge.

germeier:
Sorry, I'm to stupid for git squash. Will fix this tonight.

germeier:
replaced by #49

PR #40, opened 2015-12-12
Don't use SCRIPTDIR

digint:
When deploying your script, I found that you are using SCRIPTDIR as fall-back directory. This is very error-prone, and completely unnecessary (as BASEDIR defaults to SCRIPTDIR).

For security sake, NOTHING should be relative to SCRIPTDIR, not even the config.sh file location (I did not patch this as it would change the basic bahaviour).

Some further suggestions:
- use only one default location for config.sh (e.g. /etc/letsencrypt.sh/config.sh), as an attacker can do nasty stuff if he gets access on any of the config locations
- don't use default config if config.sh is not found, as this does most probably break things for most users

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -19,14 +19,14 @@
 # Path to license agreement (default: https://letsencrypt.org/documents/LE-SA-v1.0.1-July-27-2015.pdf)
 #LICENSE="https://letsencrypt.org/documents/LE-SA-v1.0.1-July-27-2015.pdf"
 
-# Output directory for challenge-tokens to be served by webserver or deployed in HOOK (default: $SCRIPTDIR/.acme-challenges)
-#WELLKNOWN="${SCRIPTDIR}/.acme-challenges"
-
 # Base directory for account key, generated certificates and list of domains (default: $SCRIPTDIR -- uses config directory if undefined)
 #BASEDIR=$SCRIPTDIR
 
-# Location of private account key
-#PRIVATE_KEY=${BASEDIR}/private_key.pem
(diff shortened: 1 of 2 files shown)


lukas2511:
Thank you. I agree that having this as a fallback seems like a bad idea.

About your other suggestions:

Using only one location may seem like a good idea, but on different systems that location may be somewhere completely different.

1) On a lot of BSD systems user installed software is located in `/usr/local` with configs in `/usr/local/etc`, the same is true for OSX
2) A lot of linux based systems use `/etc` as their config directory
3) A non-root user may want to use this, so loading config from user home is also a requirement
4) There are actually users managing multiple installations who want to use the config relative to the current working directory.

So yea, removing this code to load the config from multiple locations (using first found) would break some users workflow. To make it a bit more secure i changed the order of directories to check in 5120dfb5f33917dc83f6a6a4bd80d8d07e7c0722.

The other thing about the default config... yea well... i don't know. I don't think this will ever really break anything except when you are really trying to, and since it's all relative to the script directory it makes testing this software out really easy, you basically just have to create domains.txt and decide on a delivery for your challenge responses (which could just be a symlink to `.acme-challenges`) to get started.

digint:
Yeah well, grabbing config from the correct location is always a pain (as you said, every distro does it differently), and this is even more true for testing.

Having a `-f <config-file>` flag kind of solve my concerns, as all the looking-for-config-magic does not happen if I always run `letsencrypt.sh -f /path/to/myconf.sh` (e.g. in cron job).

PR #41, opened 2015-12-13
Update $WELLKNOWN default dir

metachris:
Makes it work with Nginx config in README


lukas2511:
I don't want to merge this.

The default challenge location is more like a dummy location, that can be used with the configurable hook script. If you want a different location you can set that in config.

I'd rather change the README to include instructions on where to best create the output directory, how to configure it and where to place the config snippet in nginx config.

Also if you are using this and the exact snippet i posted in README it looks like you have your private key very near to a directory that is used as webroot. You should fix that.

PR #53, opened 2015-12-16
two revoke fixes

germeier:
revoke: follow symlink and use real certificate filename so we move the file and not the cert.pem symlink
revoke: check if the given cert file exists, otherwise show error message and abort

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -452,8 +452,16 @@ command_sign_domains() {
 # Description: Revoke specified certificate
 command_revoke() {
   cert="${1}"
+  if [[ -L "${cert}" ]]; then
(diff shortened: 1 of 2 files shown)


lukas2511:
Looks like it would leave the symlink dangling around, we may want to remove that after renaming the real certificate.

germeier:
IMHO this is only a cosmetic problem
Both our webserver (can not start without a certificate) and letsencrypt.sh (will generate new cert) will behave the same with a dangling cert.pem symlink or no cert.pem file at all

lukas2511:
slightly modified version merged in c70180364b75fca01a3a299b17c5b6f67dccd26c (modified to allow links with absolute target)

PR #54, opened 2015-12-16
fix: --domain/-d is a parameter and not a command!

germeier:
$ ./letsencrypt.sh --domain test1.com --domain test2.com -c
Only one command can be executed at a time.
See help (-h) for more information.

This is obviously wrong. Fixed it.


lukas2511:
It seemed sane for me to automatically set the command to `sign_domains` with `--domain` specified, because it isn't used anywhere else, but I guess you are right.

I think this also needs to be changed a little bit, at the moment `--domain example.org --domain example.net` would result in a single certificate for both domains, would be great if we could fix that so that it behaves the same per domain parameter as it does per line in domains.txt.

germeier:
This is a design choice. I like the current approach more, e.g. only one cert will be generated when the domain parameter(s) are used

PR #58, opened 2015-12-19
Added WELLKNOWN Apache 2.4 example

Rayne:


lukas2511:
I don't really think this belongs inside this projects README file, the nginx example is just to demonstrate basic usage, I don't want to add examples for every webserver software out there.

PR #62, opened 2015-12-22
Tiny tidy up commits

erdgeist:
* Unify the use of multiple sed commands.
* Use "assing default value" expansion for defaults
* On tiny shorthand conditional

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -25,12 +25,9 @@ CONTACT_EMAIL=
 
 set_defaults() {
   # Default config variables depending on BASEDIR
-  if [[ -z "${PRIVATE_KEY}" ]]; then
-    PRIVATE_KEY="${BASEDIR}/private_key.pem"
-  fi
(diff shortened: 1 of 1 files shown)


lukas2511:
Hey, the new way of setting default parameters doesn't seem to work, i tested these changes and PRIVATE_KEY is empty as set in the beginning of the script.

Also if using `[[` (shell-builtin) works for you please use that instead of `[` (external command `/bin/[`).

erdgeist:
The [[ => [ still was left over when I thought I'd convert the script to posix shell.
The Assign default value did not assume the variable to be null instead of unset. Usually I prevent pollution by using unset. Fixed now.

lukas2511:
This pull-request has become quite dirty with all those merge commits...

Some parts of this have become unnecessary with some of the changes i just pushed (especially setting default config values), some have been merged in one of my last commits, and I'll merge some more with later cleanup commits, so i'll close this pull-request now.

PR #64, opened 2015-12-26
better use grep -q insted of output redirection

asdil12:
It's also faster as `grep -q` finishes on the first matching line.


lukas2511:
already have this in my local version of the script, will commit&push soon, thanks anyway!

PR #73, opened 2016-01-07
Also concatenate privkey.pem and cert.pem to to privcert.pem. This is needed for e.g. pure-ftpd

chrisb86:


lukas2511:
I don't know, where does this end? You'd probably also need a privfullchain.pem or something. I think using a hook for this is a better idea than making this part of the script.

mro:
indeed, the hook is how I did it for lighttpd: http://blog.mro.name/2015/12/lighttpd-letsencrypt-sh/

chrisb86:
Haven't thought about using the hook. Good idea ;)

PR #75, opened 2016-01-08
add support for dns-01 challenge

germeier:
This adds support for the new dns-01 challenge!

ATTENTION: This challenge is currently only available in staging, see: https://community.letsencrypt.org/t/dns-challenge-is-in-staging/8322

I added a example script on how to deploy a DNS challenge using nsupdate. This should be easy to port to other APIs/services.

I am uncertain if "urlbase64" is the correct way in this case or if this should be a "base64 -w0".

See also: https://tools.ietf.org/html/draft-ietf-acme-acme-01#section-7.5

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ Parameters:
  --force (-x)                     force renew of certificate even if it is longer valid than value in RENEW_DAYS
  --config (-f) path/to/config.sh  Use specified config file
  --privkey (-p) path/to/key.pem   Use specified private key instead of account key (useful for revocation)
+ --challenge (-t) http-01|dns-01  Which challenge should be used? Currently http-01 and dns-01 are supported
+ --hook (-k) path/to/hook.sh      Use specified script for hooks
 ```
 
 ### domains.txt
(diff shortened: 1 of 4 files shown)


lukas2511:
Yea good idea to implement this, but i don't like the hook being part of the repository.
Maybe we should add a section to the wiki for example hooks...

I have a larger set of changes i'll push in a minute, this pull-request will probably break while doing that, but i'll have a look at manually merging it in the next days.

germeier:
I think we want to make it users as easy as possible. I don't mind not putting the hook script into the repository but it should be documented somewhere as an example how to do things.

If you are OK with the rest of my changes I can redo the pull request sometime tonight after you pushed your changes.

lukas2511:
On the rest of the pull-request: I'd split this into two pull-requests or at least commits to have the change with the "--hook" parameter separate from the rest, also I'd change testing for the challenge-type to use a switch statement instead of multiple ifs, but otherwise it looks fine.

Also: When you do the PARAM_HOOK thingy, please make sure that it works with the new load_config method and works in command_env.

germeier:
updated with #76 and #77 so this one can be closed

PR #80, opened 2016-01-09
Add cleanup command to delete old files

code28:
Since the script keeps all the old certificates as a backup, the folders can get messy after a while. So I thought, a cleanup command would be useful. When there are a lot of different domains, there would be a lot of files to delete manually, so this command helps by checking, which certificates are used and deletes the other ones.

I was thinking about how to implement it, and I think an automated or implicit cleanup would be against the thought of having backups in case of trouble. So I came up with this extra command, so the cleanup can be started explicitly, when the certificates are working fine.

I was thinking about only deleting expired certificates, but first I wanted to start with this idea, so we can discuss on this basis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Commands:
  --revoke (-r) path/to/cert.pem   Revoke specified certificate
  --help (-h)                      Show help text
  --env (-e)                       Output configuration variables for use in other scripts
+ --cleanup (-gc)                  Remove old and not used files in certs directory
 
 Parameters:
  --domain (-d) domain.tld         Use specified domain name(s) instead of domains.txt entry (one certificate!)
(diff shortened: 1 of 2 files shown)


lukas2511:
Yea i think a bit of cleanup logic would be nice, but i would prefer to only delete expired (and unlinked) certificates, and not everything that just isn't in use anymore. Also I don't know about the way you implemented the check for used files with that array/list and stuff... I mean I guess it will work, but for some reason I just don't feel comfortable with it near my kinda important certificates.

code28:
Yes, I was thinking about it too, to only delete expired certificates. This can be added quite easily with OpenSSL's `checkend` command. That's probably safer.
The list thing was an quick and easy way of verifying, that the files aren't linked anymore, so no linked files are touched.

lukas2511:
I've implemented a solution in e60682c06c52cc52799f1eeecc178dbc3618ffdc.

My solution moves unused files to the (new) `archive` directory and ignores files that probably weren't created by the script itself so it doesn't accidentally delete user-created files.

PR #85, opened 2016-01-14
Added missing space before exit code.

jaquer:
I could've sworn the script gave me an  `exit1: command not found` error at some point, but I can't reproduce it now.

In any case, here is the world's smallest bugfix.


lukas2511:
Yea this happens if no command is given but only arguments (e.g. `./letsencrypt.sh -x`), thanks.

PR #86, opened 2016-01-15
Check if directory for LOCKFILE is writable.

petrkle:
Check if directory for LOCKFILE is writable, otherwise exit with error message.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -77,6 +77,8 @@ init_system() {
   load_config
 
   # Lockfile handling (prevents concurrent access)
+  LOCKDIR="$(dirname "${LOCKFILE}")"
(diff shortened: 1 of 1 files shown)


letsencrypt.sh
```diff
@@ -77,6 +77,8 @@ init_system() {
   load_config
 
   # Lockfile handling (prevents concurrent access)
+  LOCKDIR="$(dirname "${LOCKFILE}")"
+  [[ -w "${LOCKDIR}" ]] || _exiterr "Directory ${LOCKDIR} for LOCKFILE ${LOCKFILE} is now writable, aborting."
```
jantore: I suppose "now" should actually be "not". So close, yet so confusing. :-)

PR #100, opened 2016-01-21
Add detail as to how dns-01 hook works

davidearl:
I just got this working with my provider's DNS API, but I had to inspect the script as to how to do it, so I thought some detail might be useful in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,17 @@ This script also supports the new `dns-01`-type verification. Be aware that at t
 
 You need a hook script that deploys the challenge to your DNS server!
 
+The hook script (indicated in the config.sh file or the --hook/-k command line argument) gets four arguments: an operation name (clean_challenge, deploy_challenge, or deploy_cert) and some operands for that. For deploy_challenge $2 is the domain name for which the certificate is required, $3 is a "challenge token" (which is not needed for dns-01), and $4 is a token which needs to be inserted in a TXT record for the domain.
+
+Typically, you will need to split the subdomain name in two, the subdomain name and the domain name separately. For example, for "my.example.com", you'll need "my" and "example.com" separately. You then have to prefix "_acme-challenge." before the subdomain name, as in "_acme-challenge.my" and set a TXT record for that on the domain (e.g. "example.com") which has the value supplied in $4
+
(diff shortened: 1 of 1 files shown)


davidearl:
Yes, I know. That's why I was doing it. It all worked really nicely with my provider's DNS API: https://www.mythic-beasts.com/support/api/primary 

remarkably easy

FWIW, here's my hook script (for deploy_challenge only)

        #!/bin/bash
        if [[ "$1" = "deploy_challenge" ]]; then
            echo " ++ setting DNS for $2 to $4"
            while read DNSDOMAIN DNSAPIPASSWORD; do
                if [[ "$2" == *$DNSDOMAIN ]] ; then
                    ZONE=$(echo $2 | sed -e "s/\\.$DNSDOMAIN\$//")
                    curl -d "domain=$DNSDOMAIN" -d "password=$DNSAPIPASSWORD" -d "command=REPLACE _acme-challenge.$ZONE 300 TXT $4" "https://dnsapi.mythic-beasts.com/"
                    echo sleeping 60 seconds for change to take effect..
                    sleep 60
                    break;
                fi
            done < ./dnsapi.config.txt
        else
            echo "hook said..."
            echo "$1 $2 $3 $4"
        fi

lukas2511:
I'll merge this now, but I think hooks should actually be explained in their own readme-section. This has to be worked on.

jantore:
I think that a hook.sh.example file could also be a good way to get users up to speed quickly. I'm thinking something along the lines of this:
```bash
#!/bin/bash

function deploy_challenge {
    local ALTNAME="${1}" TOKEN_FILENAME="${2}" TOKEN_CONTENT="${3}"

    # pass
}

function clean_challenge {
    local ALTNAME="${1}" TOKEN_FILENAME="${2}" TOKEN_CONTENT="${3}"

    # pass
}

function deploy_cert {
    local DOMAIN="${1}" KEYFILE="${2}" CERTFILE="${3}" CHAINFILE="${4}"

    # pass
}

HANDLER=$1; shift; $HANDLER $@
```
I'll send a PR if you want.

lukas2511:
Yea maybe also add some comments explaining a bit when a hook gets called, that would be great.

PR #109, opened 2016-01-23
Make letsencrypt.sh zsh-compatible

Sp1l:
- Allow zsh next to bash
  - Fix small compat issues
  - Allow using a different OpenSSL binary
  - openssl sha is SHA-0, use dgst command instead

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -39,6 +39,9 @@
 # Default keysize for private keys (default: 4096)
 #KEYSIZE="4096"
 
+# Path to OpenSSL binary (default: <unset> - uses $PATH)
+#OPENSSL_BIN=/usr/local/bin/openssl
+
 # Path to openssl config file (default: <unset> - tries to figure out system default)
 #OPENSSL_CNF=
 
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -9,7 +9,7 @@ if [ "${THISSHELL}" = "/bin/sh" ]; then
   # Find compatible shell and re-exec
   for NEWSHELL in bash zsh; do
     if [ `which ${NEWSHELL}` ]; then
-      exec `which ${NEWSHELL}` "-c $0 $*"
+      exec `which ${NEWSHELL}` -c $0 $*
     fi
   done
   echo "ERROR: This script requires either bash or zsh" >&2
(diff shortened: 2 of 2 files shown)


Sp1l:
Not sure where this is failing in CI... Just tested this in a bash shell and the switching of /bin/sh to /usr/local/bin/sh was OK (FreeBSD 10.2, /bin/sh is bourne shell not bourne again)

Sp1l:
Documented in https://wiki.freebsd.org/BernardSpil/LetsEncrypt and intend to create a port for FreeBSD

nylocx:
I tested your branch and for me it goes into an endless loop with 100% CPU usage.  I'm using zsh on arch linux.

bittorf:
it think the (( ${#} )) thingy is the culprit. but i like your changes overall...

lukas2511:
I think CI also had this problem, it shows a timeout after quite a lot of time. Can't merge it as is, also didn't really look at the changes since it's not working correctly.

Sp1l:
Sorry for the poor quality code :/ Not all my local changes were in.

It now runs but there's an error unrelated to my changes:
```
STDERR:
  + ERROR: An error occurred while sending get-request to https://testca.kurz.pw/directory (Status 502)
```

Sp1l:
Tests are now OK. Any remaining comments?

lukas2511:
I'll still have to look at it, hopefully will get around to it this weekend or early next week.

I don't really like this shell-switchie-thingie, I want this to "just work" in most shells, but I see that at least some of your commits are going in the right direction.

lukas2511:
I really dislike that this contains so much more than just making it compatible with zsh. I'll manually pick out the important parts for now, the rest of the functionality may be added later.

lukas2511:
Initial zsh compatibility is now provided since da2eeda97d04817b895a6c85e72d5c5d53e32241, closing this pullrequest as I don't want that automatic shell selection and some of the other stuff was already fixed.

Sp1l:
Thanks Lukas! Works like a charm now.
Updated the FreeBSD port http://www.freshports.org/security/letsencrypt.sh/
Updated my guide https://wiki.freebsd.org/BernardSpil/LetsEncrypt

PR #110, opened 2016-01-23
read domains into temporary file and array to allow stdin to function…

nneul:
… normally in loop


lukas2511:
better workaround merged in #113

PR #113, opened 2016-01-25
pass IO descriptors to HOOK transparently

### LINKED ISSUE #101: Please add capability to allow hook to read from stdin instead of /dev/tty
Right now, the way it's written, hooks can't do normal IO then get immediate EOF.

While it's straightforward enough to run under expect/whatever, would be nice if a simple read/prompt would work.

hilbix:
closes #101 

Verified with following test-hook:

```
#!/bin/bash

echo "Args:"
for a
do
	echo "$a"
done
echo "Press return"
read
```

BTW: In contrast to the official client your script is a delight (and just works).  Thanks!

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -8,6 +8,9 @@ set -u
 set -o pipefail
 umask 077 # paranoid umask, we're creating private keys
 
+# duplicate scripts IO handles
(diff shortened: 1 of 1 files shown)


lukas2511:
I don't really like this, it looks ugly, but up to now this is the best solution so i'll merge it.

PR #115, opened 2016-01-29
Check all certificate expiration dates

crza:
Hey,
I implemented a quick certificate expiration check based on your signing function.

It currently returns 1 if a certificate is expiring and 2 when expired.


lukas2511:
Since this duplicates quite a lot of code I don't want to merge it, please find a better solution.

PR #133, opened 2016-02-12
check if curl is able to reach letsencrypt via http and https.

mitch-geht-ab:
I had the problem that http_proxy env var was not set. Because of set -e curl terminated with != 0 and the whole script terminated without error message.
So I've added 2 checks in check_dependencies() if curl is able to reach letsencrypt.org via http and https.
Another problem could be the firewall (port 80 or 443 blocked), dns or routing.


lukas2511:
I think this should be handled in-place. Curl should return a connection-problem specific return code that should just be captured in http_request()

lukas2511:
The mentioned commit should print an error if curl can't connect to the server instead of failing in silence.
This way it should also work for temporary connection problems.

PR #135, opened 2016-02-12
Updated Readme with rate limit warning + staging URL

aral:
There is currently no mention of Let’s Encrypt rate limits or the staging server. I’ve updated the readme so that the rate limit notice is prominently displayed and the necessary config setting to use the staging server is included in the instructions.


aral:
Thanks :)

PR #140, opened 2016-02-16
Add RPM spec for CentOS

gsanchietti:
Works both on CentOS 6 and CentOS 7.


lukas2511:
Sorry, I don't want any package-manager specific stuff in this repository.

PR #141, opened 2016-02-16
Handle case where no files to cleanup exist.

fancycode:
This fixes cases where `*.foo` expands to the string `*.foo` instead of an empty string if no matching files exist.

I know this is a problem with my setup, but it would be nice if letsencrypt.sh could handle such cases gracefully ;-)

Error without the change (I have a `chain.pem` in `customfolder`):
```
Moving unused file to archive directory: customfolder/chain-*.pem
mv: cannot stat '/path/to/customfolder/chain-*.pem': No such file or directory
```

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -701,6 +701,9 @@ command_cleanup() {
 
       # Loop over all files of this type
       for file in "${certdir}/${filebase}-"*".${fileext}"; do
+        # Handle case where no files match the wildcard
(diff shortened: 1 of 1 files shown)


fancycode:
Good point, I updated the pull request accordingly.

lukas2511:
Can you change it to use `-f` instead of `-e`? Then I'll merge it.

fancycode:
Changed, thanks.

PR #143, opened 2016-02-17
Use /usr/bin/env bash shebang

rumpelsepp:
On FreeBSD bash is installed in `/usr/local/bin`; this patch alters the shebang to use `/usr/bin/env bash` instead, to ensure that the shell scripts work on (particularly) the *BSDs as well.

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 ########################################################
 # This is the main config file for letsencrypt.sh      #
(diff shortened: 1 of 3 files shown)


mapreri:
though, why a shebang is needed at all in this file?

this is file is not even marked as executable, and it doesn't need to be as it is sourced by letsencrypt.sh, so why bothering with a shebang.

@rumpelsepp would be ok for FreeBSD to just remove the shebeng?  In Debian we have an automated check that warns about a non-executable having a shebeng, and we are patching the line away.

lukas2511:
@mapreri it's not really required in config, but it also doesn't hurt having it there

rumpelsepp:
@mapreri Since we keep the shebang in the main script I guess it is okay to remove it.

PR #146, opened 2016-02-18
Make path to openssl binary configurable.

fancycode:
I would like to use libressl (from https://launchpad.net/~strukturag/+archive/ubuntu/libressl) instead of openssl for everything crypto related. Therefore it would be nice to be able to configure a custom `openssl` compatible binary through `config.sh`.

### DIFF
--- a/config.sh.example
+++ b/config.sh.example
@@ -39,6 +39,9 @@
 # Default keysize for private keys (default: 4096)
 #KEYSIZE="4096"
 
+# Path to openssl binary (default: <unset> - uses "openssl" from system path)
+#OPENSSL=
+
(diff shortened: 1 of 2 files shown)


lukas2511:
I think it would be better to define an alias instead of changing every call for the openssl binary. 
You actually should be able to do that in your config file without any changes to the script itself, but making it a bit cleaner by using a config variable would totally be okay.

Sp1l:
Setting an `alias` in the config file works just fine. See https://wiki.freebsd.org/BernardSpil/LetsEncrypt#Domains_to_sign as well

lukas2511:
Thanks @Sp1l. Closing this pull-request as the changes are not really necessary.

PR #148, opened 2016-02-18
New hook deploy_certs which calls once after all certificates is issued.

onovy:
Simple use-case: Call nginx/apache reload only once after all certificates was issued.

### DIFF
--- a/hook.sh.example
+++ b/hook.sh.example
@@ -52,4 +52,27 @@ function deploy_cert {
     #   The path of the file containing the intermediate certificate(s).
 }
 
+function deploy_certs {
+    local DOMAIN_FILE="${1}"
+
+    # This hook is called after all certificates has been
+    # produced. Here you might, for instance, copy your new certificates
+    # to service-specific locations and reload the service.
+    #
+    # Parameters:
+    # - DOMAIN_FILE, with one domain per line with this format:
+    #
+    # - DOMAIN
+    #   The primary domain name, i.e. the certificate common
+    #   name (CN).
+    # - KEYFILE
+    #   The path of the file containing the private key.
+    # - CERTFILE
+    #   The path of the file containing the signed certificate.
+    # - FULLCHAINFILE
+    #   The path of the file containing the full certificate chain.
+    # - CHAINFILE
+    #   The path of the file containing the intermediate certificate(s).
+}
+
 HANDLER=$1; shift; $HANDLER $@
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -553,6 +553,8 @@ command_sign_domains() {
     _exiterr "domains.txt not found and --domain not given"
   fi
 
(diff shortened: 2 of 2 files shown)


caleb:
:+1:

This seems like a good idea. Right now I reload my nginx in the `deploy_cert` function.

the-solipsist:
@onovy Could you please add a few example lines in the hooks.sh.example file of *how* one would copy / symlink the certs, and then reload nginx / apache?  It would really help noobs like myself for whom that isn't clear.  Thanks!

lukas2511:
@onovy i think the name `DOMAIN_FILE` in example hook may be a bit confusing, and i think this is going to have problems when `BASEDIR` has spaces in its path, would be great if you could find a solution for that, but otherwise i think it's fine.

kroimon:
Would definitely be great to have such a hook.
Couldn't you just surround the paths with spaces?

lukas2511:
@kroimon No, the problem is that you may have stuff like `"/path/to/file1.pem" "/path/with spaces/file2.pem"`. As parameters to a command this will be no problem (shell will handle this for us), but written to a file this looks like `/path/to/file1.pem /path/with spaces/file2.pem`, which will basically split our second path into two wrong paths `/path/with` and `spaces/file2.pem`.

kroimon:
My idea was to include the quotes in the text file. I just don't know how easy it is to read that back from the file.

lukas2511:
@kroimon that may work, it's unlikely that quotes will appear in the actual paths, and adding commas it would become a csv file, which should be rather easy to parse. i guess i'll implement something replacing this pullrequest very soon since i need something like this myself.

kiranos:
+1

verwilst:
Hey Lukas, 

Since I'm in very urgent need for this, I was wondering if we could have a talk and try to get this thing sorted out together? I'm willing to implement it, but would rather do it in a way that helps upstream as well.

ezbik:
`DOMAIN_FILE` is not a good idea especially because of space delimited values. Basically you need to pass associative array , like this  http://pastebin.com/KaEnHqf0 . Bash 4+ has support of globally defined associative arrays so you could pass it as an argument.

Anyway, why not use this final hook only for reloading webserver, while installing certificates in `deploy_cert()` ? Then you need to pass to it nothing.

lukas2511:
I'm closing this since I think that this is actually something that can easily be handled outside of dehydrated (using hooks and wrappers) and will only result in complications (especially with per-certificate configuration which allows for different hook scripts per certificate).

If you need something like this I'd suggest doing the following:
In the `deploy_cert` hook touch a file, e.g. `/tmp/restart_apache`, then use the new `exit_hook`-hook I just added (298a7e9aafaa0b05c3f6ab6ca6538661f6b5feb8) or a wrapper around dehydrated that looks for this file, reloads Apache, and removes the file.
You can do a lot more in the `deploy_cert` hook like actually appending domains to a certain file if you need that for some reason.

PR #155, opened 2016-02-23
Add support for --keep-going in cron mode

BtbN:
Just as the title says.

The strange construct with wait is because with a regular ( ) subshell set -e does not propagate, causing the script to misbehave.

Also adds some newlines to the error function, to make the output more readable in case of errors in keep-going mode.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -712,7 +712,12 @@ command_sign_domains() {
     fi
 
     # shellcheck disable=SC2086
-    sign_domain ${line}
+    if [[ "${PARAM_KEEP_GOING:-}" = "yes" ]]; then
+      sign_domain ${line} &
+      wait $! || true
+    else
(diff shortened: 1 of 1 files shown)


lukas2511:
Hmpf... I like the idea, but this should only continue on the "too many certs" error, not in any other case. Also it should still return with an errorcode in the end.

punjabik:
I am for some keep going options too. We have hundreds of domains. I hate when some domain can end program, because challenge is invalid.
We need continue checking rest of domains.

lukas2511:
Merged in 34565c193d0360dd4abbe1e630e5cad1396e81ca and 117d5d622846e8f8efd62c112ab7e8976a0b36a7. For now I'm considering this as experimental as especially with hundreds of domains it may be easy to miss an error. Maybe adding a hook in case of an error or something similar would be nice to catch any problems.

PR #156, opened 2016-03-01
pass timestamp and key_algo to deploy hook

bax-:


bax-:
see https://github.com/lukas2511/letsencrypt.sh/issues/129

lukas2511:
Actually I think we should reduce this to just the timestamp argument and export KEY_ALGO like some other config values are exported:

```
# Export some environment variables to be used in hook script
export WELLKNOWN BASEDIR CONFIG
```

We may want to move that new export to the actual signing loop for future-compatiblity with per-cert configurations.

lukas2511:
timestamp is now a parameter of the `deploy_cert`-hook and KEY_ALGO is exported as environment variable

clydeclements:
Just one minor comment: I think you should update the `hook.sh.example` file inline with the commit above to show that `deploy_cert` now takes an additional timestamp argument. I just got bitten by this, writing a hook script with `deploy_cert` expecting 5 arguments, when 6 was required. Great software though!

lukas2511:
@clydeclements done, thanks.

PR #157, opened 2016-03-01
Add hook for unchanged certificates.

lschuermann:
In my scenario, where I automate certificate-issuance through the `letsencrypt.sh`-client, I need to also have a hook for all certificates, of which the domains haven't changed and which are still valid.

I tried to integrate this functionality, as I think more people could benefit from this feature.
As far as I am concerned, it works as expected.

I'd greatly appreciate if you could merge this pull-request.

Thanks,

Leon

### DIFF
--- a/hook.sh.example
+++ b/hook.sh.example
@@ -52,4 +52,24 @@ function deploy_cert {
     #   The path of the file containing the intermediate certificate(s).
 }
 
+function unchanged_cert {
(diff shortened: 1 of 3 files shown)


lukas2511:
Can you change it to just say `Skipping renew!`? Otherwise it's fine.

lschuermann:
Ok, I've changed the message to just say `Skipping renew!`.

PR #173, opened 2016-03-28
Using alias instead of root in cases where root does not work

sschueller:


lukas2511:
that should have been `alias` to begin with, fixed it, thanks.

PR #177, opened 2016-04-05
add hook-example howto convert certs to java keystore file

bjoern-r:
This pull request adds an example how to convert the generatet certs and key to a java keystore used by jetty and wildfly service hosts.


lukas2511:
Sorry, this doesn't belong into this repo (yet), and especially not in the main example hook template.

PR #179, opened 2016-04-07
Configuration stubs and validation hooks for use on QNAP devices

jorhett:
These files can be used as described in the README for using letsencrypt.sh on QNAP devices. Tested and working for both HTTP and DNS validation.

### DIFF
--- a/docs/qnap/README.md
+++ b/docs/qnap/README.md
@@ -1,0 +1,46 @@
+# letsencrypt.sh for QNAP units
+
+## Common Installation
+
+Alter the config.sh to enable a configuration directory
+```
+CONFIG_D=${SCRIPTDIR}/config.d
+```
+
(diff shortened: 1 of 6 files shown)


lukas2511:
Hey, thanks for this, but unfortunately I don't think this has a place inside this repository in its current for, but if you want to you can create a page on the wiki for now.

In the near future I want to add example hook scripts to this repository, and I see that most of the stuff in your pull-request is actually very generic, so maybe it would be best to wait a bit and resubmit your pull-request in parts (stunnel-deployment-example, dns instructions + wait example, ...) after I decided on a structure of examples for this repository.

jorhett:
Well the point you're missing is that with these scripts it will work "out of the box" on QNAP devices. No, there's no magic features there, but it was a platform support option.

PR #184, opened 2016-04-26
wellknown.md: Clarify WELLKNOWN variable

gboudrias:
I spent a few hours fiddling with an invalid challenge, because I had assumed the acme-challenge part was added automatically. This change should help future users not make the same mistake.


lukas2511:
Thanks. Just wondering: you wrote `must not have a trailing slash`, is this a problem you experienced? from what i see it shouldn't really make a difference, but maybe i'm forgetting something...

gboudrias:
I haven't had a problem, but I figured it's best not to be ambiguous. If you put a trailing slash it would just write to /the/path//challengefile , which in Linux doesn't matter, but it's best not to have to guess anything imo.

lukas2511:
Would you change that section to `should not have a trailing slash` and also add the missing quotes at the end of the example paths? Otherwise it's OK.

gboudrias:
Good catch. done.

PR #188, opened 2016-05-02
start of python hook example

brad-alexander:
barebones python script, gets the cert saved locally and uploads it to IAM, but doesn't do much of anything else yet.

usage:

    R53_ZONEID=Z5BBG5FOEEBBH ./letsencrypt.sh -d api.example.com -c --hook ./hook.py --challenge dns-01

### DIFF
--- a/docs/examples/hook.py.example
+++ b/docs/examples/hook.py.example
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+import boto3
+import json
+import sys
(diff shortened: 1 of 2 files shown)


lukas2511:
Hey, I wouldn't have anything against a generic hook example, but in this specific form this doesn't yet fit into this repository. Feel free to add it to the wiki!

PR #202, opened 2016-05-18
Fix issues on 2016-05-18. Remove new lines from json response.

rsauvatinet:
I found out the lets'encrypt API returned pretty json with new lines which break the sed parsing. 

This patch just remove the new lines from the json.


lukas2511:
Thanks for bringing this up, I wrote my own fix since yours didn't work under BSD and I took the chance to clean up a bit more of the incoming json.

rsauvatinet:
Great. I just found out this morning that the script wasn't working so I wrote the quick fix. Your solution is much better.
Thanks.

dl8bh:
as I did not have the time to rewrite this fix for FreeBSD sed, I replaced sed with gsed and this fix worked fine!

PR #204, opened 2016-05-21
Make location of domains.txt configurable

ymc-dabe:
> This is a follow up from: https://bugs.debian.org/824928

In the [Debian packaging of letsencrypt.sh](https://packages.debian.org/letsencrypt.sh) we would like to have the file listing the domains for which certificates should be requested at: `/etc/letsencrypt.sh/domains.txt`.
Best thing to accomplish this in our opinion is to make it possible to define `${DOMAINS_TXT}` in config.sh.

This pull request implements this feature.


mapreri:
great, thanks! ♥

PR #210, opened 2016-05-25
Make certificate output location configurable

Alexendoo:
Particularly useful if you want to set up dual certificates, you can set the config files to output into separate directories rather than using separate base directories for each. e.g. `CERTDIR="${BASEDIR}/certs/rsa"` and `CERTDIR="${BASEDIR}/certs/ecdsa"`


lukas2511:
thx

PR #218, opened 2016-05-31
letsencrypt.sh#217: FreeBSD sed doesn't parse challenges properly

### LINKED ISSUE #217: FreeBSD sed doesn't parse challenges properly
At line 381, `challenges` is set to an empty value and the script aborts due to `pipefail` at line 383.

The fix for this is to remove new lines and collapse the output into a single line.

bahamat:
Fixes #217.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -423,7 +423,7 @@ sign_csr() {
     echo " + Requesting challenge for ${altname}..."
     response="$(signed_request "${CA_NEW_AUTHZ}" '{"resource": "new-authz", "identifier": {"type": "dns", "value": "'"${altname}"'"}}' | clean_json)"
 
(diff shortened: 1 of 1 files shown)


bahamat:
After this, FreeBSD will be supported. Beyond the base system, only bash and curl are required from ports.

lukas2511:
Mh... This is weird, the first thing that happens in the clean_json function is that all newlines are removed. Also it's weird because I'm actually doing most of my local testing on OS X where most of the tools actually do come from BSD, and here everything was working fine. Guess I have to look into this a bit more...

azasypkin:
I have exactly the same issue on Ubuntu 16.04 LTS 64-bit and PR fixes issue for me as well.

bahamat:
I came back to look at this today, it seems I had been using a version before the `clean_json` function. Using your branch unmodified on FreeBSD works 👍

PR #219, opened 2016-06-01
make dependency check work for non-gnu diff

awaw:
diff on ~~*BSD~~ OpenBSD has no -v option. Replacing it with a diff between identical files works for every diff implementing -u.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -35,7 +35,7 @@ check_dependencies() {
   _sed "" < /dev/null > /dev/null 2>&1 || _exiterr "This script requires sed with support for extended (modern) regular expressions."
   command -v grep > /dev/null 2>&1 || _exiterr "This script requires grep."
   _mktemp -u > /dev/null 2>&1 || _exiterr "This script requires mktemp."
-  diff -v > /dev/null || _exiterr "This script requires diff."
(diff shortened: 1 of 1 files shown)


awaw:
Sorry, didn't try on FreeBSD. On OpenBSD I get:

% diff -v
diff: unknown option -- v
usage: diff [-abdipTtw] [-c | -e | -f | -n | -q | -u] [-I pattern] [-L label]
            file1 file2
       diff [-abdipTtw] [-I pattern] [-L label] -C number file1 file2
       diff [-abditw] [-I pattern] -D string file1 file2
       diff [-abdipTtw] [-I pattern] [-L label] -U number file1 file2
       diff [-abdiNPprsTtw] [-c | -e | -f | -n | -q | -u] [-I pattern]
            [-L label] [-S name] [-X file] [-x pattern] dir1 dir2
% which diff
/usr/bin/diff

lukas2511:
Thanks for bringing this up, I changed the check (similar to what you suggested), can you test if it is working fine for you now?

awaw:
* Lukas Schauer <notifications@github.com> [04.06.2016 04:14]:
> Thanks for bringing this up, I changed the check (similar to what you
> suggested), can you test if it is working fine for you now?

Yes, it is working fine now. Thanks.

PR #222, opened 2016-06-06
Fixed the Nginx example

WoLpH:
The current example is actually incorrect it seems :)

### DIFF
--- a/docs/wellknown.md
+++ b/docs/wellknown.md
@@ -24,8 +24,10 @@ With Nginx you'll need to add this to any of your `server`/VHost config blocks:
 ```nginx
 server {
   [...]
-  location /.well-known/acme-challenge {
-    alias /var/www/letsencrypt;
+  
+  # enable letsencrypt
+  location /.well-known/acme-challenge/ {
+      alias /var/www/letsencrypt/.acme-challenges/;
   }
   [...]
 }


lukas2511:
No, if you set `WELLKNOWN=/var/www/letsencrypt` the example should work.

WoLpH:
It hasn't worked for me on 2 separate servers. 

Can you explain how and why nginx would guess the hidden acme challenges directory? Or did my export not function properly for some reason?

lukas2511:
If you read the readme you'll find out how WELLKNOWN works, and then you'll know why the example should work.

Basically: challenge file gets saved as /var/www/letsencrypt/challengefile, gets queries at http://example.org/.well-known/acme-challenge/challengefile, nginx sees location /.well-known/acme-challenge, sets alias to /var/www/letsencrypt, resulting in nginx delivering /var/www/letsencrypt/challengefile.

WoLpH:
So somehow the script is indeed not picking up on my environment variable. Thanks for the reply

lukas2511:
@WoLpH you should set the variable in the config file, letsencrypt.sh doesn't really work with environment variables

WoLpH:
Ah, thanks. That does explain :)

Issue #178, opened 2016-04-05
Unbound variable


eknkc:
With the latest version I started receiving this error on OSX:

`./bin/le: line 27: @: unbound variable`

That line would be this:

```
_mktemp() {
  mktemp -t letsencrypt.sh-XXXXXX "${@}"
}
```

Any idea what this might be?

eknkc:
Installed newer version of bash from homebrew and this is fixed but the default osx version is not compatible I guess.

lukas2511:
Should be fixed now, can you check again?

eknkc:
Fixed, thanks!

PR #223, opened 2016-06-06
fix mktemp arg quoting

glensc:
fix for #178 broke quoting (75985c6a), arguments with spaces get exploded to different args

altho i don't have old enough bash to test with (#178 did not indicate what version he had or was it even bash), i'm pretty sure this change is safe

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -25,7 +25,7 @@ BASEDIR="${SCRIPTDIR}"
 # Create (identifiable) temporary files
 _mktemp() {
   # shellcheck disable=SC2068
-  mktemp ${@:-} "${TMPDIR:-/tmp}/letsencrypt.sh-XXXXXX"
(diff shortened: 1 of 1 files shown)


lukas2511:
I don't really understand what this should do. From what i see it would always fall back to the "$@" rule, basically changing nothing at all. Also the parameter is only used in one place (version checking) and doesn't ever get passed any spaces. Closing this as I don't think this change is useful.

glensc:
it expands to nothing with no args and expands to `"$@"` with args. thus fix for broken shell reported originally. the thing is that `$@` expands words in it (one word becomes two words), it must be used as `"$@"`. saying that "right now you evaluated only place its used" is bad practice, later you make more use-cases and will totally forget that you have broken word expansion in place. if you intend to break word expansion on purpose use `$*` not `$@`. but i really just suggest to merge this.

glensc:
also, given that you say the function always called with arguments, then the result is always `"$@"` not `$@`, there is a difference.

PR #224, opened 2016-06-07
Add Lighttpd example to wellknown.md

domenpk:


txr13:
The Lighttpd example may be confusing for some users. The alias given would require $WELLKNOWN to be set to /var/www/letsencrypt/.acme-challenges. The earlier examples for Apache and Nginx both user the given $WELLKNOWN set to /var/www/letsencrypt.

I believe the Lighttpd example should be modified as follows:
`alias.url += (
 "/.well-known/acme-challenge/" => "/var/www/letsencrypt/"
)`
This will allow users copying the example to use the same $WELLKNOWN as given for the other examples.

lukas2511:
@txr13 yup, thanks, i updated the example config in 2042b177c7c81bee5b83debb201758ba411a873d

domenpk:
Apologies, my bad. I've used /etc/ paths (I think they're mentioned in some other docs here), and then changed to /var without actually testing.

PR #225, opened 2016-06-17
Fix domain name check from certificate.

ergonlogic:
It looks like Let's Encrypt changed the way they generate certificates early this morning. There are no longer "DNS" lines, breaking the check to extract the cert name.

We (@colans & I) just switched to looking in the CN. We haven't tested with alt names yet, but will be shortly.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -697,7 +697,7 @@ command_sign_domains() {
     if [[ -e "${cert}" ]]; then
       printf " + Checking domain name(s) of existing cert..."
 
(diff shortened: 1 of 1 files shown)


memtkmcc:
How to reproduce this problem? We are using letsencrypt.sh 0.2.0 and it still works fine, like before.

lukas2511:
Can't reproduce this (tested on staging and production) and your change is incomplete and breaks the check if certificate names have changed.

colans:
Thanks for looking into it. It's very possible that we were looking at the wrong certificate as we noticed that our LE ones were sometimes being overwritten by self-signed ones in [our attempt to get this working with Aegir](https://www.drupal.org/node/2629560).  Sorry about that.

PR #226, opened 2016-06-19
Fetch WELLKNOWN from environment if available

WoLpH:
This fixes the issue I had in #222 by optionally accepting WELLKNOWN from the environment


lukas2511:
I don't think this is a good idea to just implement this for WELLKNOWN. 
Either do it for all config variables (and with a clear prefix to avoid confusion), or none.

PR #228, opened 2016-06-23
added wait time for the locking mechanism

jume-dev:
I had a requirement that i would start the script parallel with different arguments.
Therefore i wanted one incarnation to wait for the other to finish while not being required to start it again myself.


lukas2511:
But... why? Just use a script to call letsencrypt.sh and don't run it in parallel?!

jume-dev:
True, never thought about using another script :/

PR #231, opened 2016-06-29
Added option to select IP version of name to address resolution

chkhanu:
Issue is:
Name resolves to bunch of an IPv4 and IPv6 addresses, but IPv6 is unreachable from network.
Hence --ipv4 is passed to curl.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@ This file contains a log of major changes in letsencrypt.sh
 - Added option to add CSR-flag indicating OCSP stapling to be mandatory
 - Initial support for configuration on per-certificate base
 - Support for per-CA account keys and custom config for output cert directory, license, etc.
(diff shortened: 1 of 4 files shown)


lukas2511:
For what exactly is this useful? This really sounds like your network configuration is broken and this should not be fixed inside this script but rather in your network config...

chkhanu:
It's quite useful, when you have no rights to change network config; when corporative dns-server is your only option (others are just blocked) and you have almost no means to change a thing (or it will take really long time).

Ofcourse, there's an option to disable ipv6 on a kernel level, or just prioritize ipv4 addresses - but there are other applications, that rely on internal ipv6 addressing.

// would be great to have single env variable to tell resolver do only ipvX addresses per application, but nah..

Sum up: to overcome weird configurations that you can't change.

lukas2511:
Mhkay.

PR #242, opened 2016-07-09
Introduce per cert configuration directory DOMAINS_D

schutm:
As I like to keep all my configuration at the same place (/etc/letsencrypt.sh/) in this case I've created a patch to have the location of the per-cert configuration files configurable. I use this to this per-cert configuration files in /etc/letsencrypt.sh/domains.d/{domain name} in stead of /var/lib/letsencrypt.sh/certs/{domain name}.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -105,6 +105,7 @@ load_config() {
   ACCOUNTDIR=
   CHALLENGETYPE="http-01"
   CONFIG_D=
+  DOMAINS_D=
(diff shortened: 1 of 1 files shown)


lukas2511:
thx

txr13:
Out of curiosity, is it intended to leave this as an undocumented option? (There's no corresponding option in the example config file.)

schutm:
Sure see: 72f8485d6d9d6e1618fccc4d80990d5ee690a17a

PR #244, opened 2016-07-20
Make umask a bit less paranoid

Stummi:
While I understand the idea behind the 077 umask, I think its kinda problematic: Most of my service are running as their own user, and some of them seem to try to read the file after dropping the privileges, which is not possible with the files having 600 as access mask. I don't want to run all these application as lets-encrypt user or as root, and I think the most sensitive solution is to set the files group-readable.

Since on most modern Distributions a user belongs to his own distinct group this change is not a dangerous escalation, but this way I can add the users which needs to access the private keys to the lets-encrypt usergroup.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -9,7 +9,7 @@ set -e
 set -u
 set -o pipefail
 [[ -n "${ZSH_VERSION:-}" ]] && set -o SH_WORD_SPLIT && set +o FUNCTION_ARGZERO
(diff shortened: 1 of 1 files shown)


rumpelsepp:
I would either use a deploy script to move the private keys around, or make the umask configurable. I personally do not want and do not need a less paranoid umask setting.

lukas2511:
Bad idea to make it more open by default. Since config is sourced you should in theory be able to just redefine umask in your config file. Didn't try it, but it should work.

PR #247, opened 2016-08-01
Update staging.md

BernieO:
Rename private_key.pem to privkey.pem
Additionaly there is no more file called private_key.json. Maybe that comment is obsolete?
Thanks for letsencrypt.sh - works great!!


lukas2511:
The comment is obsolete since letsencrypt.sh now supports multiple accounts. Removed in ca0249c. Thanks.

PR #251, opened 2016-08-02
Update missing config message

tomchiverton:
There is no need for this message to sound so scary. letsencrypt.sh runs fine with it's defaults. I'd even consider removing this warning entirely.

### DIFF
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -126,7 +126,7 @@ load_config() {
   IP_VERSION=
 
   if [[ -z "${CONFIG:-}" ]]; then
-    echo "WARNNG: No main config file found, using defaults" >&2
+    echo "WARNING: No main config file found, using defaults" >&2
   elif [[ -f "${CONFIG}" ]]; then
     echo "# INFO: Using main config file ${CONFIG}"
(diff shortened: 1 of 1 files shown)


letsencrypt.sh
```diff
@@ -126,9 +126,7 @@ load_config() {
   IP_VERSION=
 
   if [[ -z "${CONFIG:-}" ]]; then
-    echo "#" >&2
-    echo "# !! WARNING !! No main config file found, using default config!" >&2
-    echo "#" >&2
+    echo "WARNNG: No main config file found, using defaults" >&2
```
jantore: Looks like a typo managed to sneak in.

tomchiverton:
It did ?

lukas2511:
I think this warning should exist and be big, people should at least know that there is a config, and maybe set a few important parameters so future updates don't break stuff if defaults ever change again. Making it smaller like this makes it too easy to ignore.

tomchiverton:
Your concerns are why breaking changes (to defaults) go in a change log, or documentation, no? 
Could we at least get a --quiet option or similar? After the 30th time you see it, it's pointless as you skip over it anyway.

PR #254, opened 2016-08-06
Fix default license help text in example config

leonklingele:


lukas2511:
going to change this to automatically detect license agreement url in the future, but thanks anyway

PR #261, opened 2016-08-21
Retry invalid challenge responses,  set number of retries to NUMBER

extremeshok:
Coded to your coding standards, shellchecked

This will allow for the challenge to be retied for X times when it fails (returns  invalid)

Allot of our servers are via cdn/load-balancers and they do not always sync quick enough, so will fail the first X attempts. 

This also allows your script to have resiliency with bad connections/timeouts.

````
letsencrypt.sh --cron --force --retries 10

Processing test.extremeshok.com
 + Valid till Nov 19 22:26:00 2016 GMT (Longer than 30 days). Ignoring because renew was forced!
 + Signing domains...
 + Generating private key...
 + Requesting challenge for test.extremeshok.com...
 + Responding to challenge for test.extremeshok.com...attempt: 1
 + Invalid
 + Requesting challenge for test.extremeshok.com...
 + Responding to challenge for test.extremeshok.com...attempt: 2
 + Challenge is valid!
 + Requesting certificate...
 + Checking certificate...
 + Done!
 + Creating fullchain.pem...
 + Done!
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Parameters:
  --ipv4 (-4)                      Resolve names to IPv4 addresses only
  --ipv6 (-6)                      Resolve names to IPv6 addresses only
  --domain (-d) domain.tld         Use specified domain name(s) instead of domains.txt entry (one certificate!)
+ --retries (-r) NUMBER            Retry invalid challenge responses,  set number of retries to NUMBER
  --keep-going (-g)                Keep going after encountering an error while creating/renewing multiple certificates in cron mode
  --force (-x)                     Force renew of certificate even if it is longer valid than value in RENEW_DAYS
  --no-lock (-n)                   Don't use lockfile (potentially dangerous!)
--- a/letsencrypt.sh
+++ b/letsencrypt.sh
@@ -509,17 +509,93 @@ sign_csr() {
     # shellcheck disable=SC2086
     [[ -n "${HOOK}" ]] && [[ "${HOOK_CHAIN}" != "yes" ]] && "${HOOK}" "deploy_challenge" ${deploy_args[${idx}]}
 
-    # Ask the acme-server to verify our challenge and wait until it is no longer pending
-    echo " + Responding to challenge for ${altname}..."
-    result="$(signed_request "${challenge_uris[${idx}]}" '{"resource": "challenge", "keyAuthorization": "'"${keyauth}"'"}' | clean_json)"
+    # eXtremeSHOK.com :: BEGIN
+    xres="1"
+    xl=0
+    while [ "$xres" == "1" ]; do
+      xl=$((xl+1))
+      # Ask the acme-server to verify our challenge and wait until it is no longer pending
+      if [[ -n "${PARAM_RETRIES:-}" ]]; then
+        echo " + Responding to challenge for ${altname}...attempt: $xl"
+      else
+        echo " + Responding to challenge for ${altname}..."
+      fi
 
-    reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
+      result="$(signed_request "${challenge_uris[${idx}]}" '{"resource": "challenge", "keyAuthorization": "'"${keyauth}"'"}' | clean_json)"
 
-    while [[ "${reqstatus}" = "pending" ]]; do
-      sleep 1
-      result="$(http_request get "${challenge_uris[${idx}]}")"
       reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
+
+      while [[ "${reqstatus}" = "pending" ]]; do
+        sleep 1
+        result="$(http_request get "${challenge_uris[${idx}]}")"
+        reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
+      done
+
+      if [[ "${reqstatus}" = "valid" ]]; then
+        echo " + Challenge is valid!"
+        xres="0"
+      else
+        echo " + Invalid"
+        if [[ -n "${PARAM_RETRIES:-}" ]]; then
+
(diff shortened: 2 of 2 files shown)


seefood:
Not sure how extra retries would help, this would only overload the servers. if a challange fails, it's usually a problem on the client's side.

I'll admit though I had only one odd case where retries helped, it ended up that when I turned off HTTPS redirect, it would work every time, with the redirection it worked only every 4-5 attempts (no problem with the text CA, only with the production). But this problem has since been reported, and apparently fixed.

lukas2511:
I agree with @seefood, retrying in most cases wont change that your challenge is invalid and this may only lead to more load on the letsencrypt servers, maybe even resulting in more invalid challenges because of overload ;)

If you have a special case where you _know_ that your challenge will take some time to be available you can always add a hook script which waits for the challenge to be seen before letting the script continue.

extremeshok:
Actually retries reduce the load.

Here is the specific use case.
All 1000+ servers/vm's are managed/setup by idempotent configuration management.
We run have a single node which is responsible for all ssl deployments for all the servers.

Letsencrypt does and will rate limit the number for auth attempts for a domain. 

I can understand your issue with someone setting 99999 retries, but this is virtually no different to them running the script 99999 times in a loop until the certificate has generated. I h

extremeshok:
Actually retries reduce the load.

Here is the specific use case:

All 1000+ servers/vm's are managed/setup by idempotent configuration management. Spanning multiple data centers and countries.

We run a single node aka ssl.domain.com which is responsible for all ssl deployments for all the servers. ie.  All "/.well-known/acme-challenge/" http requests from all the 1000+ servers are proxied to the ssl.domain.com server.

Under real world situations on production servers there are timeouts for the length the proxy request can take. Letsencrypt also has timeouts for the validation response. So fact is they can and will fail to validate because they timed out.

Letsencrypt does and will rate limit the number for validation attempts for a domain, so it provides no more load than the current script. Also if you look at the code, I specifically slow down the retries by adding a random 1-10 second pause between each retry.

I can understand your issue with someone setting 99999 retries, but this is virtually no different to them running the script 99999 times in a loop until the certificate has generated. The major difference is the retries will have a far lower impact on the lets encrypt servers.

Currently: If you are creating a certificate with 20 sub-domains and  only one of the 20 sub-domains fails (ie number 19) validation, you are forced to rerun the script and re-validate all the sub-domains which were already validated. Thus you increased the load 18 more times.

With my retries only the failed validation will be retried, thus you only have 1 extra validation.

If you are concerned with the user setting a limit of 99999, I have added another pull request which limits retries to 5.

In real world production all are verified by the 3rd attempt.

As to the proposed add a hook script, there is no place in the code for a hook after the json verification has been generated and before the verification has been validated.

Not to mention the limitation with the hook scripts as they cant do retries for a timeout.

lukas2511:
Normally you would do this in the deploy_challenge hook, which can be used to deploy to DNS, other Webservers or CDNs, and whatever is in use, and this is also the place for waiting until a local deployment has finished before letting the script talk to letsencrypt for verification.

Maybe parsing the error message and looking for indications for a timeout may be a possibility, also local caching for already done validations should be possible and is already something in my todo, so rerunning the script won't add additional overload for subdomains in the future.

Also I won't merge your commit in this form because you are duplicating a lot of code and you put your domain name in the middle of the script.

extremeshok:
Forgot to remove the comments.

With regards to the duplicated code, I can convert it to a do while loop, but did not want to change your logic.  Otherwise one could make a function call.

With regards to the deploy hook, ill give you a real world example.. Servers located in mainland china have around a 300msec lag when connection to our usa datacenter. Even if one uses the hook to check the file is accessible. Letsencrypt can and does timeout/fail. 

Another issue iis with geolocated servers and loadbalanced/round robin dns. Its very difficult to predict the path/route letsencrypt will use to communicate. One sub-domain can be 18 locations. These all then need to proxy the request back.

shoeper:
> Even if one uses the hook to check the file is accessible. Letsencrypt can and does timeout/fail.

I don't understand this. You could e.g. use curl with with 100 retries and the code would block the check until curl terminates. You could also call sleep 30 to give your infrastructure 30 seconds to deploy the challenges.

extremeshok:
@shoeper the validation is done via letscrypts own network. A curl would be done on the machine running the script aka your network.

So even if the script can access the url, does not mean letsencrypt will be able to.

We do not deploy, everything is done via nginx proxy and redirections. Letsencrypt, the proxy and the ssl server can all timeout or fail a single request.

extremeshok:
BTW: this is not theoretical, from actual production usage over many months. This has been the only reliable and consistent method.

If you have 1000+ domains,100+ servers in more than 10 countries, then you will understand why this is required.

adduxa:
By the way once I got 50x the error code on the validation request, so at least curl with some retries looks like a good idea.

lukas2511:
@extremeshok sorry, i still think that it is your or your systems responsiblity to know when the deployment of files to your servers has happened. if you have no way of knowing if your deployments finished something else in your network needs fixing. i agree that in some situations there should be retries (e.g. on timeouts), but if you have a 404 because your file isn't available yet that is your problem and you should handle that before letting the script (and with that the verification) continue.

extremeshok:
@lukas2511 

The files are instantly deployed, as they are on the ssl.domain.com server I dont think you understood what I wrote.

> We do not deploy, everything is done via nginx proxy and redirections. 

>Letsencrypt, the proxy and the ssl server can all timeout or fail a single request.

lukas2511:
@extremeshok ok, then we are on the same point, should retry on timeouts. there needs to be code that checks for the actual error that happened (parsing error message and reading curls return code), i don't want the script to retry on user errors.

PR #264, opened 2016-08-26
filter commonName

bborbe:
I had a problem using letencrypt.sh under pfsense (FreeBsd). Finally i fixed it with a grep -v on commonName.

My error message:

```
...
 + Requesting challenge for example.com...
 + Responding to challenge for example.com...
 + Challenge is valid!
 + Requesting certificate...
  + ERROR: An error occurred while sending post-request to https://acme-v01.api.letsencrypt.org/acme/new-cert (Status 400)

Details:
{
  "type": "urn:acme:error:malformed",
  "detail": "Error creating new cert :: policy forbids issuing for: common name (eg, your name)",
  "status": 400
}
```

openssl version -d

> OPENSSLDIR: "/etc/ssl"

cat /etc/ssl/openssl.cnf |grep commonName

> commonName                      = Common Name (eg, YOUR name)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,10 @@ This file contains a log of major changes in letsencrypt.sh
 
 ## [x.x.x] - xxxx-xx-xx
 ## Changed
+- ...
+
+## [0.3.0] - 2016-09-07
+## Changed
 - Config is now named `config` instead of `config.sh`!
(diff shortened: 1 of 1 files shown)


lukas2511:
In this case copy and edit your openssl config and set OPENSSL_CNF to the new location, this should not be handled by dehydrated.

PR #269, opened 2016-09-07
Adding the invalid_challenge and request_failure hooks

nkovacne:
To us it is very useful to know when a challenge has failed (we use the dns-01 method and sometimes it fails because TXT DNS record has not propagated yet), so I created the "invalid_challenge" hook. This way we are aware and can act accordingly.


lukas2511:
Can you please have a look at what you have done with docs/hook_chain.md, looks like you duplicated some of the text.

nkovacne:
I just added a hook chain example that involves an invalid challenge, so the new hook gets called. If you think it is improper I will remove that.
```
+ + Challenge is invalid!
￼ +HOOK: invalid_challenge lukas.im the_reason
```

lukas2511:
The diff shows that you readded the whole block with your change, but the old one is still there, so now the text is duplicated.

nkovacne:
Oops, you're right, sorry, bad place for that example. I guess the example in examples/hook.sh should be enough.

nkovacne:
I also added another hook: request_failure, in case the curl call should fail for some reason. The http_request() function already has a hook call (clean_challenge), but it is not exclusive for the request failure as clean_challenge is also called from other places where the request was ok.

Up until now we've got some 50X codes, so it's good to us to know when this happens and, in case failure extends in time, to be aware that something is not working good (probably on the client's server side).

lukas2511:
The commits with this pull request got way to messed up. Can you please clean up your commits and also please create a separate pullrequest for the second hook? In this current state I don't want to merge it.

nkovacne:
Sure, I'll tidy up code and create separate PRs.

PR #272, opened 2016-09-14
Debian packaging for dehydrated

allardhoeve:
I wrote some quick Debian packaging. This works, but there are a couple issues:

1. Running it as an unpriviliged user requires you to either modify the config file or chown `/var/lib/dehydrated`.
2. Docs on how to properly get stuff working are missing, but it is a little out of my scope right now.

To build on ubuntu:

```bash
sudo apt-get install build-essential devscripts debhelper fakeroot
fakeroot debian/rules binary
```


lukas2511:
There already is a package (even in the debian repository!) https://packages.debian.org/jessie-backports/letsencrypt.sh
I contacted one of the package maintainers about renaming the package, and at some point I'll probably merge the debian control/rules/whatever files into this repository.

Thanks anyway.

allardhoeve:
Woops. Well then :)

PR #277, opened 2016-09-15
add --create-dirs flag

vdloo:
create necessary directories that do not exist with --create-dirs or -cd

```
$ dehydrated -c
ERROR: WELLKNOWN directory doesn't exist, please create /data/web/public/.well-known/acme-challenge and set appropriate permissions.
```

```
$ dehydrated -c --create-dirs
Processing example.com
 + Valid till Dec 14 09:53:00 2016 GMT (Longer than 30 days). Skipping renew!
```

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -59,6 +59,7 @@ store_configvars() {
   __OPENSSL_CNF="${OPENSSL_CNF}"
   __RENEW_DAYS="${RENEW_DAYS}"
   __IP_VERSION="${IP_VERSION}"
+  __CREATE_DIRS="${CREATE_DIRS}"
 }
 
 reset_configvars() {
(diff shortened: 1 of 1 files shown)


lukas2511:
I don't really think I want this in here.

Having to create the directory manually (or by Ansible/Puppet/whatever) isn't really much work, doing it automatically just adds unnecessary code for a one-off operation, and it adds another option making readability of the usage instructions a bit more complicated...

In cases where you have a dynamic challenges-directory you probably have something for generating the call to begin with, let that create the directories.

vdloo:
OK no problem, that's a valid consideration. We put it in [our fork](https://github.com/ByteInternet/dehydrated/pull/1).

PR #278, opened 2016-09-15
Adding the invalid_challenge hook

nkovacne:
This hook is called if the challenge response has failed, so domain owners can be aware and act accordingly.

### DIFF
--- a/docs/examples/hook.sh
+++ b/docs/examples/hook.sh
@@ -74,4 +74,18 @@ function unchanged_cert {
     #   The path of the file containing the intermediate certificate(s).
 }
 
+function invalid_challenge {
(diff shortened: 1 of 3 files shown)


docs/examples/hook.sh
```diff
@@ -74,4 +74,18 @@ function unchanged_cert {
     #   The path of the file containing the intermediate certificate(s).
 }
 
+function invalid_challenge {
```
el-dge: hook.sh is in /bin/sh not in bash. (function don't exist)
invalid_challenge() {

nkovacne: Right, I just pushed another PR fixing it. Thanks.

PR #280, opened 2016-09-17
fix some README typos

benjixx:


lukas2511:
Thanks, I'll let these slip in with the next bigger change to the README.

PR #283, opened 2016-09-20
Fix request new certificate if authorization valid

dennisschuerholz:
I just had an issue after switching from dns based challenges to http (see [serverfault](https://serverfault.com/questions/804030/letsencrypt-certificate-renewal-switching-from-dns-to-http) and [le-community](https://community.letsencrypt.org/t/certificate-renewal-switching-from-dns-to-http/20030?u=dennisschuerholz)) so I made this quick fix which allowed me to use dehydrated again.

The change in Line 464(old)/468(new) corresponds to the "bigger" answer as mentioned in the serverfault post. Indeed it may not longer be necessary as I implemented a "quick escape" before extracting the challenges but I would leave it there - just in case.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -463,6 +463,8 @@ sign_csr() {
     if [[ $response == *"\"status\": \"valid\""* ]]; then
       already_valid[${idx}]=true
       continue
+    else
+      already_valid[${idx}]=false
     fi
 
     challenges="$(printf '%s\n' "${response}" | sed -n 's/.*\("challenges":[^\[]*\[.*}\s*]\).*/\1/p')"
--- a/dehydrated
+++ b/dehydrated
@@ -450,9 +450,9 @@ sign_csr() {
 
   local idx=0
   if [[ -n "${ZSH_VERSION:-}" ]]; then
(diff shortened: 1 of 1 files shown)


dehydrated
```diff
@@ -450,18 +450,24 @@ sign_csr() {
...
-    local -a challenge_uris challenge_tokens keyauths deploy_args
+    local -a challenge_uris challenge_tokens keyauths deploy_args already_valid
   fi
 
   # Request challenges
   for altname in ${altnames}; do
     # Ask the acme-server for new challenge token and extract them from the resulting json block
     echo " + Requesting challenge for ${altname}..."
     response="$(signed_request "${CA_NEW_AUTHZ}" '{"resource": "new-authz", "identifier": {"type": "dns", "value": "'"${altname}"'"}}' | clean_json)"
+    if [[ $response == *"\"status\": \"valid\""* ]]; then
```
sth: There are several `"status": ...` fields in the response, but probably it works out ok if any of them is `valid`. I assume the toplevel `status` would be the one we really are interested in.

`$reponse` should be quoted in case it contains spaces.

lukas2511:
Closing this in favor of #293, thanks anyway!

PR #291, opened 2016-10-05
Output in PKCS#12 format

ahnooie:
Added PKCS#12 (PFX) cert output for importing into windows (tested on Server 2012 R2 with IIS 8.5).  Can specify a password to protect the key in config/param, otherwise defaults to blank password.


txr13:
PFX output should be optional, not mandatory.

I'm currently using dehydrated on a single, master certificate server for all domains and certificates across our company. Some certificates are required in PEM format, some are required in PFX format, and some (in both formats) have password requirements on the private key. I've written a separate script to process the raw PEM files to add passwords or convert to PFX (with or without passwords) before outputting them to the required target.

Given that this is a bash script, it seems more likely to be run in a Linux environment, where PEM files are pretty standard (in my experience). Having PFX output as an option in dehydrated wouldn't be a bad thing, but I wouldn't like it to be mandatory--many servers will have no use for PFX files. If PFX output was made a configurable option, though, it could be turned on using a per-certificate config file, making it a more flexible (and useful!) option to have.

ahnooie:
Good suggestion @txr13 with the latest commit PFX creation is optional (disabled by default), also it can be turned on/off and password can be set on a per certificate basis using the method described in docs/per-certificate-config.md

lukas2511:
Sorry, but I don't really see the point in doing this inside of dehydrated, this could easily be done by a hook-script, all required parameters should be available at that point, or is there anything I'm missing about this? Also please keep your commits cleaner, in the current state I would not want to merge this...

ahnooie:
Thanks for your consideration, for now I'll use my fork but will look into replacing it with a hook if I get the chance.

Question (since this is my first PR on Github).   I figured you would squash on merge.  Is it standard practice to rebase commits to clean them up before doing a PR?

PR #286, opened 2016-09-26
Create Alpine based Dockerfile

valentin2105:
Hi, 

I wrote this small Dockerfile (27.81Mb) based on Alpine Linux. 

Do you want use it ? 

https://github.com/valentin2105/dehydrated-docker
https://hub.docker.com/r/valentinnc/dehydrated/

PR #292, opened 2016-10-05
Dockerfile: Add Docker support

jmreicha:
Adds the necessary dependencies and files to a Dockerfile so the dehydrated script can be run inside a container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN apk add --update \
     openssl \
     && rm -rf /var/cache/apk/*
 
+WORKDIR dehydrated
 COPY . .
 
 ENTRYPOINT ["bash", "dehydrated"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM alpine:latest
+
+RUN apk add --update \
+    bash \
+    curl \
+    openssl \
+    && rm -rf /var/cache/apk/*
+
+COPY . .
+
+ENTRYPOINT ["bash", "dehydrated"]


Rayne:
See https://github.com/lukas2511/dehydrated/pull/286

jmreicha:
I added a [wiki page](https://github.com/lukas2511/dehydrated/wiki/Docker-support).

I'm curious why nobody wants Docker support?

extremeshok:
Because its utterly useless. Obviously you drank too much of the docker cool-aid.

Dehydrate is a simple script, there is no need to use it as a docker.  It also relies on a webserver or a dnsserver to fullfill the request, so your docker image is invalid for practical points.

jmreicha:
That was incredibly rude.  I was just asking a question.

extremeshok:
@jmreicha that was me being nice, sorry if you were offended.

lukas2511:
Just to clarify: I don't have nothing specific against a Dockerfile (although in this case I really don't see the point of running a simple script like this inside a container), I just don't want any kind of packaging files inside this repository for now.

If I'd start merging stuff like this I'd end up with a repository with Dockerfiles, Vagrantfiles, puppet playbooks, chef cookbooks, Debian package files, rpm package files, PKGBUILD for archlinux, ebuild file for gentoo, lxc example config, freebsd package files, nixos package files, ... and I don't want that. Especially because I'd have to update most of them every time I do a new release.

jmreicha:
@lukas2511 That makes sense, I appreciate the explanation.

PR #293, opened 2016-10-08
Skip challenge for already validated domains

sth:
For domains that are already validated we don't need to fulfill further challenges.

If a challenge request for a domain returns `"status": "valid"` no further validation is required and further processing is skipped. Skipping this further processing also avoids parsing problems when the challenge response contains additional fields, for example describing a previous successful DNS challenge.

The new variable `${challenge_altnames}` is used to keep track of the names that are not validated yet and still need do the challenge. When later iterating over this possibly empty array a `:0` substring expansion has to be used to avoid bash complaining about an unbound variable (it treats empty arrays as unset).

This solves the same problem as pull request #283, hopefully in a more robust way.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -305,6 +305,13 @@ get_json_string_value() {
   sed -n "${filter}"
 }
 
+rm_json_arrays() {
+  local filter
+  filter='s/\[[^][]*\]/null/g'
+  # remove three levels of nested arrays
+  sed -e "${filter}" -e "${filter}" -e "${filter}"
+}
+
 # OpenSSL writes to stderr/stdout even when there are no errors. So just
 # display the output if the exit code was != 0 to simplify debugging.
 _openssl() {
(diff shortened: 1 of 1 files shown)


lukas2511:
Thanks for this, looks fine to me but I want to do some tests on FreeBSD and ZSH, just to make sure it works, but this will definitively get merged!

lukas2511:
Just gonna merge this now, can still fix it later anyway if there are any problems, thanks!

whereisaaron:
This patch causes errors whenever a domain is already validated, because the 'continue' causes the initialization of several variables to be skipped.

```
 + Already validated
./dehydrated: line 441: challenge_altnames[@]: unbound variable
```

Commenting out the 'continue' is a workaround, but of course defeats the goal of the patch.

See additional reports of errors cause by this patch in issue #320

PR #298, opened 2016-10-16
Keep code quality high with shellcheck

szepeviktor:
See https://github.com/koalaman/shellcheck/wiki/TravisCI

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -3,6 +3,8 @@
 # Fail early
 set -eu -o pipefail
 
+NGROK_URL="https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-stable-linux-amd64.zip"
+
 # Check if running in CI environment
 if [[ ! "${CI:-false}" == "true" ]]; then
   echo "ERROR: Not running in CI environment!"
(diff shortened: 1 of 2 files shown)


szepeviktor:
Please consider fixing these shellcheck warnings

https://travis-ci.org/lukas2511/dehydrated/builds/168084127#L269

test.sh
```diff
@@ -3,6 +3,8 @@
 # Fail early
 set -eu -o pipefail
 
+NGROK_URL="https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-stable-linux-amd64.zip"
```
leonklingele: Why was this changed?

leonklingele: (I'm referring to the URL)

szepeviktor:
See https://ngrok.com/download


-- 
+36204242498
Ezen a készüléken sok az elütés. 
Elnézést!

test.sh
mrwacky42: That's where the [ngrok Download](https://ngrok.com/download) page points. [Equinox.io](https://equinox.io/) seems to be some sort of Go project hosting.

lukas2511:
I agree that shellcheck is useful and cleaning the code should be done, but I don't want any tiny false-positive resulting in tests failing, so I won't merge this into the test environment.

PR #311, opened 2016-11-24
Fixing "$1 not set" for HANDLER var

a5a351e7:
Just cloned your repository, configured config and all necessary files. 
Started a test with `dehydrated -f /etc/dehydrated/config -c` and got the following

```
root@domain:/etc/dehydrated# /opt/dehydrated/dehydrated -f /etc/dehydrated/config -c
# INFO: Using additional config file /etc/dehydrated/hook.sh
/etc/dehydrated/hook.sh: Zeile 77: $1 ist nicht gesetzt.
```

So it seems that the script is called without parameter and got this exception. So I added in my pull request a simple if-else-statement!

### DIFF
--- a/docs/examples/hook.sh
+++ b/docs/examples/hook.sh
@@ -74,5 +74,7 @@ unchanged_cert() {
     #   The path of the file containing the intermediate certificate(s).
 }
 
-HANDLER="$1"; shift
-"$HANDLER" "$@"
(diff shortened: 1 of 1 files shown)


lukas2511:
I don't think it's worth merging this, that part of the example hook script should be written a lot cleaner anyway (like using a case-statement instead of just executing whatever is given as first parameter), thanks anyway.

PR #318, opened 2016-12-14
Simplify CSR (config) generation

TCM2:
openssl doesn't accept - as an input filename, but there is still /dev/stdin you can use to feed a config directly.

Used positional parameters instead of $altnames to simplify the creation of the subjectAltName= line.

Just a proposal from my own script.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -620,21 +620,22 @@ sign_domain() {
 
   # Generate signing request config and the actual signing request
   echo " + Generating signing request..."
-  SAN=""
-  for altname in ${altnames}; do
-    SAN+="DNS:${altname}, "
-  done
-  SAN="${SAN%%, }"
-  local tmp_openssl_cnf
-  tmp_openssl_cnf="$(_mktemp)"
-  cat "${OPENSSL_CNF}" > "${tmp_openssl_cnf}"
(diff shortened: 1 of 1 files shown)


dehydrated
```diff
@@ -620,21 +620,22 @@ sign_domain() {
...
-  if [ "${OCSP_MUST_STAPLE}" = "yes" ]; then
-    printf "\n1.3.6.1.5.5.7.1.24=DER:30:03:02:01:05" >> "${tmp_openssl_cnf}"
-  fi
-  openssl req -new -sha256 -key "${CERTDIR}/${domain}/${privkey}" -out "${CERTDIR}/${domain}/cert-${timestamp}.csr" -subj "/CN=${domain}/" -reqexts SAN -config "${tmp_openssl_cnf}"
-  rm -f "${tmp_openssl_cnf}"
-
+  (
+    cat "${OPENSSL_CNF}"
+    
+    printf "[SAN]\nsubjectAltName="
```
TCM2: For consistency, this could be echo -n instead of printf. Same below.

lukas2511:
I can't recall what exactly but I'm pretty sure that I tried this and I had a problem with it, also temporary files don't really hurt anybody so I'm leaving it as it is right now.

PR #325, opened 2016-12-28
Fix: bash to sh notation on invalid_challenge() hook

nkovacne:
The hook for invalid_challenge still had old notation from bash, I converted the bash notation to sh notation.


lukas2511:
Commited that in 81eecedc5af058dc69b14331088aa2f69c60c554 because your pull-request got poisoned by other commits. Please use branches for pull-requests in the future so this doesn't happen again :)

Issue #335, opened 2017-01-05
challenge tokens are created in wrong directory for freeBSD


tekbasse:
challenge tokens are created in WELLKNOWN dir instead of WELLKNOWN/.well-known/acme-challenge/

for bash under freeBSD 10.3  See #334 

After commenting out line 545:
` [[ "${CHALLENGETYPE}" = "http-01" ]] && rm -f "${WELLKNOWN}/${challenge_token}"`

The files are found in WELLKNOWN dir.

lukas2511:
This is not a bug, this is a feature: https://github.com/lukas2511/dehydrated/blob/master/docs/wellknown.md

tekbasse:
I call it a documentation bug. Poorly explained.

lukas2511:
In that document I state multiple times how files are stored, and there even is example configuration for probably every popular webserver, I'm really not sure what else I can do.

If you have a suggestion on how to make the documentation better feel free to send in a pull-request.

tekbasse:
Okay. Will do.

tekbasse:
The acme standard requires the file to be served from /.well-known/acme-challenge/

> The path at which the resource is provisioned is comprised of the fixed prefix
".well-known/acme-challenge/", followed by the "token" value in the challenge.
The value of the resource MUST be the ASCII representation of the key
authorization.

`.well-known/acme-challenge/evaGxfADs6pSRb2LAv9IZf17Dt3juxGJ-PCt92wr-oA`

from: [https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md](url)

So, WELLKNOWN needs to include this fixed value at the end.

The pull request will reflect this description.

cheers, Ben

PR #336, opened 2017-01-05
per issue #335, updating documentation with slight change to config file

tekbasse:
Added SERVERROOT, which is common terminology to help web developers quickly identify the intended value of SERVERROOT and consequently, the value of WELLKNOWN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,8 @@
 
 *Note: This project was renamed from letsencrypt.sh because the original name was violating Let's Encrypts trademark policy. I know that this results in quite a lot of installations failing but I didn't have a choice... For now there is a wrapper script for compatibility with old config locations and symlinks, but it will be removed in a few weeks.*
 
-This is a client for signing certificates with an ACME-server (currently only provided by Let's Encrypt) implemented as a relatively simple bash-script.
-
-It uses the `openssl` utility for everything related to actually handling keys and certificates, so you need to have that installed.
-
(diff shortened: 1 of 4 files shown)


lukas2511:
From your introduction of the concept of a "SERVERROOT" I see that you are making a few assumptions about the system this script is running on:

- Has a Webserver running, reachable from the outside
- Has a single root from which all websites are served (e.g. `/var/www`)
- The script is able to write to a directory that can be accessed from the outside

This might not be true in a lot of cases, there are a lot of setups with distributed servers and central certificate management out there, using hook scripts to actually deploy challenges and certificates, "WELLKNOWN" in those cases is not a directory that is directly accessible over the internet but more like a working directory for different systems.

Introducing the concept of a "SERVERROOT" conflicts with more complex setups, so this is not good for this script, also it introduces a completely unused variable to the config file, which doesn't really make sense.
I also read your modifications to the documentation and I think they make it even more confusing, that doesn't really help.

Thank you for trying, but I'm closing this since I don't think these changes are a good idea.

tekbasse:
Not so, Lukas.

These changes:

1. bridge a missing gap in the documentation for users that are not operating complex systems;

2. fix some disjointed topics and ambiguous language and grammar; and

3. updates a description of WELLKNOWN per changes in draft ACME standards. 

These changes do not adversely affect any complex practice of WELLKNOWN in any way. 

SERVERROOT can be ignored for any that don't expect it.

You are free to ignore this pull request, or incorporate some or all of the changes under your own authorship.

best wishes,
Ben

PR #339, opened 2017-01-10
Debianized

Vitexus:
Hallo, 
there is files required for building basic debian package for dehydrated

Thank for your work!


ezbik:
@Vitexus According to cron file, it is executed under root? Unprivileged user should be used instead, I believe.

ezbik:
@Vitexus /var/www/letsencrypt  should be changed to /var/lib/letsencrypt, because it may interfere with user's other files in /var/www

lukas2511:
dehydrated is already in the debian package tree (stretch and sid) and I won't add package files to this repository, thanks anyway.

leonklingele:
Why not use the packaging which is already available in the Debian repos? https://packages.debian.org/search?keywords=letsencrypt.sh

Vitexus:
Sorry man my mistake :)

Issue #154, opened 2016-02-23
"keep-going" option for cron mode


BtbN:
Because of the plain amount of sub-domains I am using I keep running into the rate-limit a lot of the times.
A lot of the other domains would still renew successfully, but the script in --cron mode aborts after the first cert fails to renew.

It would be extremely useful to have an option to make it continue with the other certs if one of them fails.

lukas2511:
I think this would be great in combination with a `failed_sign` hook (open for better names), which would make things like #160 easier.

Cyborgscode:
the "keep going" features is required for many real world situations. 

- a domain does no longer point to the server which created the cert.
the script does not know this, so it tries to renew the cert, which fails and it stops => no renew for the other domainnames.

Solution: 
1) enable dns checks for the domainnames, where the IN A has to point to an ip in the output of "ip addr" .
2) enable a admin notification emailadress to get an email if a fail on the creation or renew of a cert happens.
3) keep going with the next cert.
4) keep track of the amount of fails, incase LE has a failrate limit.

Lets be honest, your script is now in use by a lot of hosters, as it's the only easy to integrate solution on the market. You properbly will get a lot a those stability request in the near future.

lukas2511:
@Cyborgscode

1) we may add a pre-sign hook so you can do checks in a hook before anything else happens, this won't become a part of letsencrypt itself.
2) related to #160, there will be more hooks making this easier to implement
3) yes, by now i agree that continuing would be a good idea in most cases
4) i'm not sure about this one, may be implemented at a later point

PR #343, opened 2017-01-26
Changed a fatal to non-fatal to be able to continue parsing domains.t…

einar:
…xt if a host experiences network issues. This could be done so much prettier but we needed something fastish as a single host in our domains.txt with reachability issues caused no further processing to happen. Hope this feature(ish) can get added even if this patch is ignored; then we can go back to using the main repo. Thanks.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -581,7 +581,10 @@ sign_csr() {
       done
     fi
 
-    _exiterr "Challenge is invalid! (returned: ${reqstatus}) (result: ${result})"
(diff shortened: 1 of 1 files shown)


txr13:
Why does `--keep-going` not meet your requirements? (See issue #154 and the referenced commit.) I'd think this is exactly what you need for your use case...

einar:
It does, I should have looked more closely at the code, sorry. We have an old version running and I looked at open issues and couldnt find it solved there so I simply thought it had been missed. We'll be using -g instead, thank you! Sorry for wasting your time. :)

PR #350, opened 2017-02-04
Don't require WELLKNOWN for `register` command

jmandel:


jmandel:
This behavior was tripping me up and forced me to work around it - any reason not to accept this change?

lukas2511:
@jmandel i did implement your change with aed4272e973bb691b66ffa59fb9b90b952803c01 but i had to change it a little bit because it was causing syntax errors, sorry for not giving any comment on it

jmandel:
Thanks!

PR #352, opened 2017-02-04
Add "command_version" to display current version with --version or -v

muffl0n:
```
> ./dehydrated | grep version
 --version (-v)                   Show version
```
```
> ./dehydrated --version
dehydrated 0.3.2
```
```
> ./dehydrated -v
dehydrated 0.3.2
```

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -4,6 +4,7 @@
 # Source: https://github.com/lukas2511/dehydrated
 #
 # This script is licensed under The MIT License (see LICENSE for more information).
+VERSION=0.3.2
 
 set -e
 set -u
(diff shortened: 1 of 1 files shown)


lukas2511:
Mh yes, was thinking about this too, but I see a problem with people using versions directly from the git repository, I'd have to constantly update the version number in the script to make it "correct" in those cases...

muffl0n:
We could leave the VERSION-variable empty in master. In command_version we could check for that and do a "git rev-parse --short HEAD" if no version is specified.
When doing a release the VERSION-variable has to be filled though.

jbergstroem:
I miss this too. My take would be to bite the bullet and add `VERSION=0.4.0` to dehydrated. Could be committed as part of your changelog updates.

lukas2511:
I've implemented this with some additional information (OS and used software versions) in f2b589430c55b01c03e382bd58e82ae6c961d175. Thanks for the suggestion.

PR #354, opened 2017-02-12
Support dynamic domains list

raffaellod:
I added support for a new config option, HOOK_DOMAINS (defaults to no), which allows to get the domains list from the hook instead of domains.txt.
I updated all the documentation I could find to mention this new option and the corresponding new hook command.

My use case is running dehydrated on a server that is running a name server (NSD in my case), so I already have a list of domains and host names in the zone files for the server, and I don’t want the burden of maintaining the two separate files.
I can publish the 10-line script I wrote that parses the zone files and outputs in a domains.txt-like format, if you can provide me with a location or file where that would be fitting.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -56,6 +56,7 @@ store_configvars() {
   __HOOK="${HOOK}"
   __WELLKNOWN="${WELLKNOWN}"
   __HOOK_CHAIN="${HOOK_CHAIN}"
+  __HOOK_DOMAINS="${HOOK_DOMAINS}"
   __OPENSSL_CNF="${OPENSSL_CNF}"
   __RENEW_DAYS="${RENEW_DAYS}"
   __IP_VERSION="${IP_VERSION}"
@@ -70,6 +71,7 @@ reset_configvars() {
   HOOK="${__HOOK}"
   WELLKNOWN="${__WELLKNOWN}"
   HOOK_CHAIN="${__HOOK_CHAIN}"
+  HOOK_DOMAINS="${__HOOK_DOMAINS}"
   OPENSSL_CNF="${__OPENSSL_CNF}"
   RENEW_DAYS="${__RENEW_DAYS}"
   IP_VERSION="${__IP_VERSION}"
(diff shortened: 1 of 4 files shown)


jbergstroem:
Nice! It makes it even more obvious that passing options per-certificate/domain other than in certificate `$output_dir` should be a thing! Perhaps expand each line in domains.txt env style; CONFIG_FOO=bar domain.com www.domain.com

raffaellod:
I finished my DNS zone file hook. If you’re interested in seeing how it integrates (and needs) this pull request, you can find it at https://github.com/raffaellod/dehydrated-zonefiles-hook .

lukas2511:
dehydrated has a ton of command line options so you can easily write wrapper scripts and run it for specific domains with different hooks and whatever you like, completely without using any domains.txt

Also I think this change is a lot more complicated than it really needs to be... I mean your whole code could be replaced by this wrapper:

```bash
#!/bin/sh
generate-domains-txt.sh > /etc/dehydrated/domains.txt
dehydrated -c
```

I really think this is a lot more beautiful than adding new config parameters, new checks, new methods and everything to dehydrated just to call `generate-domains-txt.sh`...

Even if you are arguing about running it inside a cronjob and not wanting another executable script file on your system for whatever reason, you could easily write that inside your crontab.

If you really wrap your code around dehydrated you can even define what hook script to use, which challenge type to use (e.g. dns-01), and even a path to a custom config just for that particular call.

I might reconsider this if you give me a really good reason why this has to be a hook, but from what I see it just doesn't make sense so I'm closing this pullrequest, sorry.

jbergstroem:
@lukas2511 thanks for your feedback. 

@raffaellod we should probably extend the wiki with "best practices".

raffaellod:
@lukas2511 I did think of the option you suggest, after I was finished and I could see what my work amounted to (split between this PR and the separate hook project). I chose to submit the pull request anyway because my solution doesn’t involve temporary files, and it has arguably a more “built-in” appeal to it. These are really the only things this PR would improve over what you suggest as alternative. Obviously this is not “a really good reason”, so I understand your decision.

I think I’ll submit a separate PR to avoid the temp files anyway, since that’s IMO a nice side effect of what I wrote. It will overall make dehydrated shrink by a few lines; I think that’s desirable. If you think otherwise, please let me know anytime, so I won’t bother.

@jbergstroem I’d be happy to update the wiki, and if possible add a reference to the hook script I wrote, after I’ll give it a README (which I was waiting the end of this PR to write, since its usage changes without this PR obviously).

PR #357, opened 2017-02-19
dont strip  when cwd is /

chasebolt:
if dehydrated is in `/dehyrdrated` with no config file, using args only, it fails as it strips `/` from `BASEDIR` leaving it as an empty string.


lukas2511:
Merged in 0dcf94dd3d6d97387479595c0c2f8bb6c277cff8

Issue #363, opened 2017-02-28
Support for storing challenge to a file


aryanraj:
Challenges for each alternate name is fetched and then challenges are deployed. Due to some network error, the script exits and the challenge tokens are lost. This leads to exceeding the rate limit.

lukas2511:
This shouldn't happen anymore since dehydrated now checks if a domain is already verified.

PR #364, opened 2017-02-28
Support for storing challenge to a file

aryanraj:
In reference to #363. After each request, challenges are stored in a local file. In the case of failure, the challenges are read from the file in the subsequent run.


lukas2511:
This shouldn't be necessary anymore since dehydrated now checks if domains are already verified.

PR #367, opened 2017-03-19
Enable alternate domain file location

Serafean:
Usecase:Per service domains.txt.
For instance ejabberd and nginx need certificates for the same domain, but with different parameters. See systemd-usage.md how the -D parameter helps streamlining dehydrated usage.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1160,6 +1160,8 @@ main() {
          fi
         ;;
 
+      # PARAM_Usage: --domains-txt (-d) domains.txt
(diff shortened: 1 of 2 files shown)


luishdez:
Hoping that this get merge asap , if you have installed dehydrated globally it's 100% necessary

lukas2511:
Since other parameters may as well be changed I'd suggest just using a separate configuration file. Should be easy to implement, if you really want to you could have a main config that is sourced by other configs just for this, a few tiny files which will replace a ton of not really required functionality.

PR #369, opened 2017-03-20
Fix typos

jwilk:
Found using [mwic](http://jwilk.net/software/mwic).


lukas2511:
Thanks for the suggestions, I commited this manually (see d1f215b652aa843d5104c5a1201acfd866db383e) since there were some conflicts and it was easier to just do the changes by hand instead of solving for merge conflicts...

PR #373, opened 2017-03-28
Add MSYS support

haddoncd:
A couple of minor changes to not break under MSYS (eg. git bash on Windows). Feel free to reject this if you don't care.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -299,7 +299,7 @@ init_system() {
 
 # Different sed version for different os types...
 _sed() {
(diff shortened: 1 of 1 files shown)


lukas2511:
Hu, `The subject starts with a /, so MSYS will assume it's a path` they seem to have done a great job with that 🙈

Since this doesn't seem to intefere with anything else I have merged it in ba31a505d29d115ff4d268dc43d23e21ab8f5006.

PR #377, opened 2017-03-31
use existing CSR in cron mode.

yxzzx:
enable dehydrated to use an existing CSR when running in cron mode. Add the variable
USE_EXISTING_CSR=<path to csr>
to ${CERTDIR}/${domain}/config

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -73,6 +73,7 @@ reset_configvars() {
   OPENSSL_CNF="${__OPENSSL_CNF}"
   RENEW_DAYS="${__RENEW_DAYS}"
   IP_VERSION="${__IP_VERSION}"
+  USE_EXISTING_CSR=
 }
 
 # verify configuration values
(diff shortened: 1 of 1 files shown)


lukas2511:
For what exactly would this be useful? The cron-mode is not really designed for custom CSRs, and I don't see the reason for reusing an old CSR. Feel free to reopen if there is a important use-case for this.

yxzzx:
I have a couple of HP systems with a management console called "iLO". The devices will issue a CSR on request and then you can upload the certificate generated with this CSR into the device. It is not possible to upload a certificate/key pair. But I still would like to update the certificates automatically.

Gnarfoz:
While I'm not using LE for this, yet, other hardware vendors' integrated management solutions (like Dell's iDRAC, Fujitsu's iRMC, etc) behave similarly. They give you a CSR to sign, no way to push own keys. (Just my 0.02€.)

lukas2511:
I'll be implementing this as a hook so it can be used a bit more dynamically. Closing this pull-request in favor of #475.

Issue #378, opened 2017-04-06
Allow quiet/silent operation


j4Hu:
- add new --quiet/--silent argument, so program report only errors (cron).
- errors to stderr

temp sollution:
dehydrated  --cron | egrep -v '^(# INFO:| \+ Checking| \+ Valid till|Processing)'
tnx

helmo:
Most errors already went to STDERR, two remainig ones are fixed in #381.

To get quiet behaviour I suggest redirecting the STDOUT to a logfile instead of grepping it away.

`dehydrated --cron >> /var/log/dehydrated.log`

cultcom:
j4Hu is right, normal operation without any change should be quit for cron. Another logfile which needs rotation and stuff is too much in my opinion.
But what I want to know is when there actually was a renewal or if that renewal failed.

dafyddt:
I'm glad I found this script as it's a much nicer solution for my purpose. I was going to replace certbot with this script across all my servers until I saw this issue.

I agree that a non-verbose cron operation would be more useful. One workaround is to run with a wrapper - a nice one is Cronic http://habilis.net/cronic/

lukas2511:
I'd recommend using a wrapper like cronic (which also seems to be just a shell-script).

Rewriting the logging code is not something I'll be working on any time soon since for me using a wrapper works perfectly and on most systems I actually just let cron send me the whole output because it only runs once a month anyway (and could be run even less often if it spams too much).

If anybody really wants this feature feel free to implement it and open a pullrequest. If i think that it's good I'll merge it.

PR #381, opened 2017-04-17
Error logging

helmo:
While looking into #378  I noticed a few lines were still printed to stdout instead of stderr. And two lines where redirection happens twice.


helmo:
The build failure seems to be a network hickup tests run OK locally. Please restart them.

lukas2511:
Merged, thanks.

PR #386, opened 2017-05-01
DNS-01 challenge response only after DNS propagation

Contik:
In a situation where a registered domain has more than one nameserver, all of which are equally weighted and thus equally likely to be the one answering Let's Enrypt's query for a proper DNS TXT record Let's Encrypt sometimes queries too quickly and doesn't find the TXT record it needs. It gets a reply from a nameserver that hasn't had the TXT record propagated to it yet.

This change uses `dig` (which is now a dependency) to check all of a registered domain's authoritative nameservers for presence of the correct DNS TXT record. Challenge response is only triggered when all of a registered domain's authoritative nameservers reply with the correct record.

With this change `dehydrated` by default begins at a 10-second wait time after the first failed nameserver check and doubles that amount for each consecutive check: nameserver checks start 10, then 20, then 40 seconds apart and so on. Default maximum wait time is 3,600 seconds or 1 hour after which `dehydrated` does its challenge response anyway which may or may not succeed (same as now).

_Anecdotal data point: For registrar INWX with a set of 5 equally weighted nameservers on a registered domain the average wait time for propagation to all 5 of them is 4mins 30s ± a few seconds after which `dehydrated`'s challenge response always succeeds whereas prior to this change success was more of a gamble._

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -36,6 +36,7 @@ check_dependencies() {
   command -v grep > /dev/null 2>&1 || _exiterr "This script requires grep."
   command -v mktemp > /dev/null 2>&1 || _exiterr "This script requires mktemp."
   command -v diff > /dev/null 2>&1 || _exiterr "This script requires diff."
+  command -v dig > /dev/null 2>&1 || _exiterr "This script requires dig."
 
   # curl returns with an error code in some ancient versions so we have to catch that
   set +e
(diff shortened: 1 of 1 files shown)


txr13:
I believe this should be implemented in the hook script, not in dehydrated itself. The API assumes the hook script will do the needful, and this includes hooks that are called to deploy other challenges.

Let's take a theoretical case of an HTTP challenge that has to be deployed and propagated across a CDN for reliable retrieval. In this case, the hook script would also need to conduct its own checks before returning control to dehydrated.

One example of a hook script which does include a propagation check can be found at https://github.com/bennettp123/dehydrated-email-notify-hook.

Issue #239, opened 2016-07-04
Feature: Account update


gertcuykens:
Commands:
--account (-a) Update contact information

https://ietf-wg-acme.github.io/acme/#registration

For example, to update the contact information in the above registration, the client could send the following request:

```
POST /acme/reg/asdf HTTP/1.1
Host: example.com

/* BEGIN JWS-signed request body */
{
  "resource": "reg",
  "contact": [
    "mailto:certificates@example.com",
    "tel:+12125551212"
  ],
}
/* END JWS-signed request body */
```

lukas2511:
This feature is now implemented, thanks to @lukavia!

PR #388, opened 2017-05-04
Added new feature Update registration contact #239

lukavia:


lukas2511:
Definitively needs some more work, but is a good starting point.

Why do you check the value of `$FAILED` just after setting it, this will always be true, seems completely unnecessary and makes the code harder to read. 
Some parameters need to be in quotes (especially for that `mv` command at the end). 
Errors should be redirected to stderr, ideally using `_exiterr`. 
For temporary files (NEW_ACCOUNT_KEY_JSON) please use `_mktemp`. 
Maybe writing a backup file for the old json would be a good idea, in case something went wrong that wasn't catched anywhere. 
A bit more information for the user on what is going on would be nice, as far as I can see this code only really logs anything on errors, but a "Updating account information..." message or something similar would be nice.

lukavia:
Re: `$FAILED` You are correct. I've copy-paste from the Init function and left it for polishing later. Fixed now. 
I've quoted parameters where appropriate.
Used _exitcode for exiting with error
Used _mktemp for temporary file
Make a backup file if the new one differs from the old one and don't touch anything otherwise
Added information to the user what is happening

lukavia:
Hey, do you need anything else in this pull request?

lukas2511:
@lukavia hey, sorry, i know it's been a while, i'm a bit busy but i'll look into your changes soon (hopefully later this week)

lukas2511:
Thanks for the feature, I just merged it in commit ec1599e3b66582da471a70e0fb02fecdeaae4b31. Sorry that this took so long...

Issue #379, opened 2017-04-12
Patch for configuring openssl command


s5410s:
I installed openssl and libressl together, and I wrote patch for switching between them

New configuration keyword: OPENSSL_CMD="openssl"

[patch.zip](https://github.com/lukas2511/dehydrated/files/916415/patch.zip)

O2Graphics:
Hello,

I'm not sure it's the best way to contribute on Github with a patch as a .zip file.
Anyway, to use LibreSSL, you don't need to add a new paramater, you can just add:
`alias openssl='/usr/local/bin/openssl'` in your dehydrated's config file.

Best Regards.

PR #393, opened 2017-05-28
Allow to use a different openssl binary than system default openssl

ixs:
There might be a different openssl binary installed which should be
preferred by dehydrated over the regular openssl-named one.

Add a OPENSSL_BIN config item to support overriding.


O2Graphics:
Seems similar to: https://github.com/lukas2511/dehydrated/issues/379

Issue #284, opened 2016-09-21
letsencrypt.sh/dehydrated: line 80: unexpected argument `(' to conditional binary operator


kayyazka:
Hi, 

I have problem on this line 

letsencrypt.sh/dehydrated: line 80: unexpected argument `(' to conditional binary operator

And i change argument with add "" in (http-01|dns-01)

from :
[[ "${CHALLENGETYPE}" =~ (http-01|dns-01) ]] || _exiterr "Unknown challenge type ${CHALLENGETYPE}... can not continue."

to :
[[ "${CHALLENGETYPE}" =~ "(http-01|dns-01)" ]] || _exiterr "Unknown challenge type ${CHALLENGETYPE}... can not continue."

after that. i run  ./letsencrypt.sh --config /etc/pki/config.sh --cron
and i get error 

ERROR: Challenge is invalid! (returned: invalid) (result: {
  "type": "http-01",
  "status": "invalid",
  "error": {
    "type": "urn:acme:error:unauthorized",

Additional information:
cat /proc/version
Linux version 2.6.18-8.el5 (brewbuilder@ls20-bc2-14.build.redhat.com) (gcc version 4.1.1 20070105 (Red Hat 4.1.1-52)) #1 SMP Fri Jan 26 14:15:21 EST 2007

uname -a
Linux  2.6.18-8.el5 #1 SMP Fri Jan 26 14:15:21 EST 2007 i686 i686 i386 GNU/Linux

cat /etc/*release
Red Hat Enterprise Linux Server release 5 (Tikanga)

what should i do to fix it?

Thank you in advance

kayyazka:
I have update new script but still error like it

lukas2511:
Uff... support for linux distributions from the stone age...

You could try deleting the accounts directory, maybe there just is a broken account key, but I have a feeling that if this one line results in a problem some others might too...

I'll leave this open for now, but keeping compatibility with bash versions from probably over ten years ago is definitively not something I want to spend too much time on.

lukas2511:
Not going to fix this, sorry.
I'd suggest trying to update your version of Bash, or better: Update your whole OS, does that thing even get security updates anymore?...

PR #394, opened 2017-05-28
Support older bash releases

ixs:
Bash 3.0 and others seem to have serious issues running dehydrated.
https://github.com/lukas2511/dehydrated/issues/284 tracks the
problem but got closed as it seems too hard to support "stone age"
distributions.

Turns out it is actually only a three line change. ;-)

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -77,14 +77,14 @@ reset_configvars() {
 
 # verify configuration values
 verify_config() {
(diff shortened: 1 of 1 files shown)


lukas2511:
Merged in 9ea75e7cfbcdfe30be2934fd293012c66677b7db, but please be aware that I won't really care if this breaks in the future.

ixs:
No worries. I would normally not have bothered but the patch was super simple. As long as it's easy to maintain, I am happy to send another patch when it breaks. If it is too hard, we can always just sunset older bash support... It has been nearly a decade after all.

PR #395, opened 2017-05-28
Ability to provide extra curl options

ixs:
In some situations it might be necessary to pass extra commands to
the curl binary, e.g. proxy authentication credentials.

Adds the CURL_OPTS config option.


lukas2511:
I'm not sure how well this will work, but I guess you have tested it and since it doesn't seem to interfere when it's unused I merged this in bd57777c6206ce991739b2681d09fee2b55b4afe.

PR #400, opened 2017-06-08
add cert alias, fixing #396

### LINKED ISSUE #396: Allow multiple keys/certs for a domain
Using domains.txt, afaik it is not possible to create multiple keys/certs for a domain or set of domains at the same time. For example, if I want to have a RSA and a ECDSA key for a domain, I cannot use domains.txt but have to do my own scripting around.

typingArtist:
This change adds a feature where each line in domains.txt can be prefixed by "--alias alias-name", where the given alias-name is used instead of the primary (first) domain name on that line, for looking up the configuration and for storing the certificate. With that change it is possible to generate multiple certificates with the same primary domain name and differentiate the configuration and output path via the alias-name. One use case is for generating both an RSA and an ECDSA key/cert combination. This change fixes #396.

[x] tested manually
[x] added simple test case to `test.sh`
[x] added documentation updates

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,7 +3,7 @@ This file contains a log of major changes in dehydrated
 
 ## [x.x.x] - xxxx-xx-xx
 ## Changed
-- ...
+- domains.txt and the commandline "--domain" allows an alias name to be given by leading "--alias alias-name"
 
 ## [0.4.0] - 2017-02-05
 ## Changed
--- a/dehydrated
+++ b/dehydrated
@@ -678,16 +678,16 @@ walk_chain() {
 
 # Create certificate for domain(s)
 sign_domain() {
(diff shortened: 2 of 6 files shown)


typingArtist:
Note: This seems a little related, but in the end quite different, to #367.

lukas2511:
I see your problem, I like the idea, but I really don't like this implementation.

Also, like I posted as response to #367:

> Since other parameters may as well be changed I'd suggest just using a separate configuration file. Should be easy to implement, if you really want to you could have a main config that is sourced by other configs just for this, a few tiny files which will replace a ton of not really required functionality.

In your case you would have a config for ecdsa certs, one for rsa certs, probably different output directories for the certificates, every other parameter could just be shared.

PR #412, opened 2017-07-12
add host for OCSP request

it-can:
This seems to work for me, don't know if it is correct

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1045,7 +1045,7 @@ command_sign_domains() {
       if [[ "${update_ocsp}" = "yes" ]]; then
         echo " + Updating OCSP stapling file"
         ocsp_timestamp="$(date +%s)"
-        "${OPENSSL}" ocsp -no_nonce -issuer "${chain}" -verify_other "${chain}" -cert "${cert}" -respout "${CERTDIR}/${domain}/ocsp-${ocsp_timestamp}.der" -url "${ocsp_url}" > /dev/null 2>&1
+        "${OPENSSL}" ocsp -no_nonce -issuer "${chain}" -verify_other "${chain}" -cert "${cert}" -respout "${CERTDIR}/${domain}/ocsp-${ocsp_timestamp}.der" -url "${ocsp_url}" -header "HOST" "ocsp.int-x3.letsencrypt.org" > /dev/null 2>&1
         ln -sf "${CERTDIR}/${domain}/ocsp-${ocsp_timestamp}.der" "${CERTDIR}/${domain}/ocsp.der"
       fi
     fi


txr13:
That looks like an awfully LE-specific value, to be hard-coded like that.

Bearing in mind that this needs to work across _any_ ACME-compliant CA, is there a more dynamic way to detect and set this, if it's required?

it-can:
I don't know...

txr13:
What does ${ocsp_url} evaluate to? Is there a way to extract the desired host header from that?

it-can:
http://ocsp.int-x3.letsencrypt.org

txr13:
I can't write the code while I'm mobile, but what about using cut, splitting on the forward slashes, and just keeping the third field? I think that should extract just the host...

lukas2511:
@it-can mh for me this works fine without an additional host argument... can you please post the output of `dehydrated --version`? Just interested on why this fails, not meaning that this will not be fixed.

Otherwise (as @txr13 stated) this pull-request is LE-specific and it also will probably only work with that specific sub-CA and will break in the near(ish) future. I'll rewrite this to take the existing parameter, closing this pullrequest.

it-can:
```
[root@server dehydrated]# ./dehydrated --version
Dehydrated by Lukas Schauer
https://dehydrated.de

Dehydrated version: 0.4.0
GIT-Revision: 4f3bd3e9567e97e963cb80141c3ba9147605360c

OS: \S
Used software:
 bash: 4.2.46(1)-release
 sed: sed (GNU sed) 4.2.2
 curl: curl 7.29.0
 awk: GNU Awk 4.0.2
 grep: grep (GNU grep) 2.20
 mktemp: mktemp (GNU coreutils) 8.22
 diff: diff (GNU diffutils) 3.3
 openssl: OpenSSL 1.0.1e-fips 11 Feb 2013
```

I am getting this error without the HOST parameter:

```
openssl ocsp -no_nonce -issuer "123.chain" -verify_other "123.chain" -cert "../certs/test.com/cert.pem" -respout "test.der" -url "http://ocsp.int-x3.letsencrypt.org"
Error querying OCSP responsder
140201268983712:error:27076072:OCSP routines:PARSE_HTTP_LINE1:server response error:ocsp_ht.c:250:Code=400,Reason=Bad Request
```

lukas2511:
Please see e339b28159a86720898ad60bb90207ca68977192, this should fix your issue.

it-can:
Doesn't seem to work...

it-can:
yup, removed the ocsp.der file and called dehydrated, no new file...

it-can:
mmm it seems to work when I remove the domain directory...

lukas2511:
I can reproduce the issue now, it seems that openssl has changed the syntax for the header parameter from version 1.0.x to 1.1.x... Meh... :disappointed:

it-can:
aaah, yeah that makes sense...

lukas2511:
Please see 2b76d038d374acc9b4237da7e224be9e48ac2439, does this work for you?

it-can:
yes that seems to work! One question, everytime I call the dehydrated script will it generate a new ocsp.der file?

it-can:
aah there is a age check in there I see

evaryont:
So to summarize, older versions of openssl did not include the Host header for OCSP requests, and on top of that, the syntax of specifying headers changed. (Between versions 1.0 and 1.1)

That's interesting.

it-can:
I am running Centos7

```
openssl version
OpenSSL 1.0.1e-fips 11 Feb 2013
```

PR #417, opened 2017-08-01
Added hardlink support, changed spacing in help

Ghent:
Hi,

I have a few cases where hardlinking is better suited, I thought you might find this a useful patch to add.  Thanks!

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -61,6 +61,7 @@ store_configvars() {
   __OPENSSL_CNF="${OPENSSL_CNF}"
   __RENEW_DAYS="${RENEW_DAYS}"
   __IP_VERSION="${IP_VERSION}"
+  __HARDLINK="${HARDLINK}"
 }
 
 reset_configvars() {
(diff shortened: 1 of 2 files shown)


lukas2511:
Can you give an example for when this would be really useful? I can't think of anything... closing this for now, thanks anyway.

Ghent:
It's useful inside of volumized containers (such as dockers) when the main files live outside the volume.  For instance, having multiple dockers needing the same certificates/keys, they can live on the host and be linked into the docker volume, rather than having to have a copy inside each docker.

Also useful for permissions purposes, again using something like SNI and needing multiple copies of the same certificate/key pair but needing different applications to be able to access them, ie: nginx user, postfix user, ftpd user.  These programs typically drop root privileges and do not always have permission to something like `/etc/ssl`.  Being able to hard link the cert/key to a location suitable for the application and then changing the permissions on the linked files to readable by the user that application runs as.

These are things that soft links (symlinks) cannot do.

lukas2511:
Yea but in that case wouldn't it be best to use a deployment hook anyway, and in there you could always just use readlink to resolve the real file and create a hard-link yourself, in any way you like, with correct permissions and everything... Still don't really see a reason in adding this to dehydrated itself.

Ghent:
Sounds good, thank you!

PR #424, opened 2017-08-18
configurable OPENSSL variable for each certificate.

nanorkyo:
I want to sign any certificates from custom CSRs generated by appliance equipments, which are HPE iLO4, Infoblox and others.  Almost #377 is same, but not different  approch.

I decided to make OPENSSL variable configurable for each certificate, like `OPENSSL=APPLIANCE_openssl` in ${certdir}/config, not only global.

If **_APPLIANCE_**_openssl called `req -new ...` from dehydrated, it call Generate CSR API to its own appliance equipment like following.

> ilo4_openssl () {
>    if [[ x"${1:-}${2:-}" = x"req-new" ]]; then
>        shift 2
>        while [ "$#" -gt 0 ]; do
>            case "${1}" in
>            -out)
>                ilo4_GenerateCSR $(basename $(dirname ${2})) ${2}
>                ;;
>            *)
>                ;;
>            esac
>            shift
>        done
>    else
>        openssl "${@}"
>    fi
> }

I'm tested on following environment.

-  FreeBSD 11.1-R
- dehydrated v0.4.0 from ports and customed.
- target appliances: HPE iLO4, Infoblox TE-1410(Grid Manager)

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -58,6 +58,7 @@ store_configvars() {
   __HOOK="${HOOK}"
   __WELLKNOWN="${WELLKNOWN}"
   __HOOK_CHAIN="${HOOK_CHAIN}"
+  __OPENSSL="${OPENSSL}"
   __OPENSSL_CNF="${OPENSSL_CNF}"
   __RENEW_DAYS="${RENEW_DAYS}"
   __IP_VERSION="${IP_VERSION}"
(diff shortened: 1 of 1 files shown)


lukas2511:
Don't really want to do this as dehydrated uses openssl in some other places too (e.g. for generating the signatures for signed requests to the CA), and this may be confusing and in some cases might even reduce security by overriding key sizes or by using a vendors openssl version with broken randomness or something similar.

If you want to do this please just use separate config files and make sure that your openssl binary or wrapper script is generating secure keys.

The CSR functionality is something I'll be working on a bit more in the future, also with new hook scripts for generating CSRs etc., so this modification would become unnecessary anyway.

Thanks anyway for the suggestion.

Issue #426, opened 2017-08-24
version information on FreeBSD


welwood08:
I noticed there's a couple of very minor things that don't work in the version information on my system (FreeBSD 10.3 jail on FreeNAS 9.10 host):
```bash
$ ./dehydrated --version
#
# !! WARNING !! No main config file found, using default config!
#
Dehydrated by Lukas Schauer
https://dehydrated.de

Dehydrated version: 0.4.0
GIT-Revision: 0be0ab083f290afbc757b8388a80df458ddfd33c

cat: /etc/issue: No such file or directory
OS:
Used software:
 bash: 4.4.12(2)-release
 sed: sed: illegal option -- -
 curl: curl 7.55.1
```
At this point the program is hanging waiting for Ctrl+D, after that it resumes:
```bash
 awk: awk: unknown option -W ignored
 grep: grep (GNU grep) 2.5.1-FreeBSD
 mktemp: mktemp: illegal option -- -
 diff: diff (GNU diffutils) 2.8.7
 openssl: OpenSSL 1.0.2l  25 May 2017
```
Not super important in my opinion, although I was momentarily confused by the hang while awk was waiting for input.

lukas2511:
This should be fixed with ce9b42d8ad5bc225b7eeeaf47df0690b5d6b54e3 and bc20ec79f3a0edc30b53ed532cdf3e17169f12a6.

PR #429, opened 2017-09-07
fix issue #426 - version info on FreeBSD

marcin-gryszkalis:
on FreeBSD base tools don't have --version. Also there's no /etc/issue.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -842,15 +842,19 @@ command_version() {
   revision="$(cd "${SCRIPTDIR}"; git rev-parse HEAD 2>/dev/null || echo "unknown")"
   echo "GIT-Revision: ${revision}"
   echo ""
-  echo "OS: $(cat /etc/issue | grep -v ^$ | head -n1 | _sed 's/\\(r|n|l) .*//g')"
+  [[ -f /etc/issue ]] && echo "OS: $(cat /etc/issue | grep -v ^$ | head -n1 | _sed 's/\\(r|n|l) .*//g')"
   echo "Used software:"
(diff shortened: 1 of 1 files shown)


dehydrated
```diff
@@ -842,15 +842,19 @@ command_version() {
   revision="$(cd "${SCRIPTDIR}"; git rev-parse HEAD 2>/dev/null || echo "unknown")"
   echo "GIT-Revision: ${revision}"
   echo ""
-  echo "OS: $(cat /etc/issue | grep -v ^$ | head -n1 | _sed 's/\\(r|n|l) .*//g')"
+  [[ -f /etc/issue ]] && echo "OS: $(cat /etc/issue | grep -v ^$ | head -n1 | _sed 's/\\(r|n|l) .*//g')"
```
shello: It's possible to get the equivalent to `/etc/issue` on FreeBSD by running `uname -sr`, which prints both the name of the operating system (`-s`) and the release (`-r`) to stdout. Both these switches are [part of the POSIX `uname`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/uname.html) and exist on all currently [supported releases](https://www.freebsd.org/security/security.html#sup) (and probably other *BSD systems too).

marcin-gryszkalis: I think /etc/issue is not really something to rely on. eg. on my old gentoo box it contains
'This is \n.\O (\s \m \r) \t
so original line seems to be pretty useless - I just wanted to suppress error from cat...

lukas2511:
Merged in ce9b42d8ad5bc225b7eeeaf47df0690b5d6b54e3.
I also added the freebsd version like @shello suggested (see bc20ec79f3a0edc30b53ed532cdf3e17169f12a6).

Thanks.

PR #434, opened 2017-09-21
Add optional user and group configuration

danimo:


dehydrated
```diff
@@ -165,6 +167,16 @@ load_config() {
    done
   fi
 
+  # Check if we are running & are allowed to run as root
+  if [[ ! -z "$DEHYDRATED_USER" && $EUID == 0 ]]; then
+    if [ ! -z "$DEHYDRATED_GROUP" ]; then
+        group="-g $DEHYDRATED_GROUP"
+    fi
+    echo "# INFO: Running $0 as $DEHYDRATED_USER"
+    su -c $0 $group "$DEHYDRATED_USER"
```
Rayne: `$0` and `$group` aren't escaped. I assume that `$group` wasn't escaped to provide the `-g` flag and the (safe) group name as individual arguments.

danimo: $group is on purpose, and I saw little point in escaping $0, but I probably should, for good measure.

danimo:
New iteration, now using sudo

lukas2511:
Thanks for your work. I've rewritten the code so that it actually checks the current UID/GID and doesn't assume non-root to be the correct user. It now also enforces the group parameter.

PR #435, opened 2017-09-22
Add man page

danimo:


txr13:
Previous PRs for Docker files (ex. #292) and Debian packaging scripts (ex. #339) have been refused. Owner's comment at https://github.com/lukas2511/dehydrated/pull/292#issuecomment-252056436 is clear that he doesn't intend to add any packaging stuff at this time, largely because of the need to keep them updated as well.

I realize this isn't strictly a "packaging" file, but I'm wondering if the need to keep it updated might also lead to its non-inclusion. It seems like a very well-written man page... perhaps it could be submitted to package maintainers for Debian/RedHat/etc.?

danimo:
I am the package maintainer for SUSE, and I am in touch with @lukas2511 about this. If there is need to simplify maintenance (e.g. by sharing generating the man page from markdown by means of generator frameworks), I'm happy to help there as well, but it seemed that at this time, pure man/troff, seemed the simplest way.

lukas2511:
Looks like it needs to be polished a bit more, there are some typos and the text itself seems like it could be improved, but merged as an initial draft in 3ec54e7e0fb7a19abf4830efb21a4a9d5292a09a. Thanks!

PR #443, opened 2017-10-08
#396 alias take 2

typingArtist:
This is the second attempt to adding the alias feature. Thanks to Dirk’s input, the feature has now a more concise configuration syntax and allows specifying of multiple aliases, i.e. multiple similar certificates for the same set of domains on one line in `domains.txt`.

- [x] implemented
- [x] added docs
- [ ] extensively tested

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,7 @@ This file contains a log of major changes in dehydrated
 - Certificate chain is now cached (CHAINCACHE)
 - OpenSSL binary path is now configurable (OPENSSL)
 - Cleanup now also moves revoked certificates
+- domains.txt and the commandline "--domain" allows one or more alias names to be specified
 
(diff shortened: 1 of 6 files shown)


lukas2511:
Hey, I'm really sorry for the long delay without any response from my side...

I've finally took the time to take a look at your work, and decided that I didn't really like the implementation of multiple aliases...

Your first commit was merged with a few changes (I wanted to keep it as low-profile as possible), but I did my own implementation of parsing the (single) alias parameter from `domains.txt`.
As a bonus I added a `--alias` switch to be used in combination with `--domain`.

I'm closing this pull request now since the original wanted feature has been implemented. If you'd like the possibility of multiple aliases maybe we can come up with a simpler solution than what you had implemented.

Thank you for your suggestion and your time.

PR #444, opened 2017-10-13
fix ocsp.der symlink

Exagone313:
The symbolic link pointing to the latest `ocsp.der` is incorrect, check how `ln -s` is used to create other links (those other work).


lukas2511:
merged in 742c0ad176f173118d8d4b19d19cec962741f99b

PR #453, opened 2017-11-30
retry post if 500 and echo warning to stdout

fragpit:
Hello! I forked this repo a year ago and faced many time out errors. I fixed it in my private repo like in pull request. Now I forked it again  and found that bug is still there. So I think my input would help to mitigate 500 errors when issuing certificates for over 300 domains with cloudflare DNS hook.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -398,8 +398,15 @@ http_request() {
     statuscode="$(curl ${ip_version:-} ${CURL_OPTS} -s -w "%{http_code}" -o "${tempcont}" "${2}")"
     curlret="${?}"
   elif [[ "${1}" = "post" ]]; then
-    statuscode="$(curl ${ip_version:-} ${CURL_OPTS} -s -w "%{http_code}" -o "${tempcont}" "${2}" -d "${3}")"
-    curlret="${?}"
+    statuscode=1
+    while [[ ! "${statuscode:0:1}" = "2" ]]; do
+        statuscode="$(curl ${ip_version:-} ${CURL_OPTS} -s -w "%{http_code}" -o "${tempcont}" "${2}" -d "${3}")"
+        curlret="${?}"
+        if [[ ! "${statuscode:0:1}" = "2" ]]; then
+            echo "WARNING: Statuscode is: ${statuscode}, retrying..." >&2
+            sleep 2
+        fi
+    done
   else
     set -e
     _exiterr "Unknown request method: ${1}"


lukas2511:
Mhmhm... I'm still not sure if I really like this. Normally when the CA throws an error it's because people didn't configure the client correctly (wrong wellknown path or didn't wait for dns refresh) or because the CA is under heavy load...

In the first case I really dislike retries, if it's not configured correctly there really is no point and it may even end up in an infinite loop (first a few retries, followed by api limit errors).

If the CA is under heavy load I'm really not sure if automatic retries are a good idea either... maybe with a bigger and staggered delay and with a maximum number of tries (first 5 seconds delay, if still error go up to 10 seconds, max out at 15 seconds and exit with error), this wouldn't put that much stress on the CA and it would limit the amount of generated api calls in case of a misconfiguration.

lewellyn:
I'm in general agreement with @lukas2511 here. As LE is meant to be automated, and any automated server should be automating it with a service somehow *and* ideally should be monitored for failures, this PR seems somewhat counterproductive.

If there's an error (500 or otherwise), there's a problem that needs resolving. If it's a timeout, it's something you need to work with your provider on or resolve another way; retries will only give the illusion that all is well. If you can't reliably reach LE, your users likely cannot reliably reach you. Retrying until it works keeps one from knowing there's a problem.

With systemd, you can just use `Restart=on-failure` provided `dehydrated` exits non-zero for these 
rare cases where it is actually desired. (If it doesn't exit non-zero, that's another PR that I'll gladly cheerlead for! 🥂 ) Other init systems have similar functionality, aside from extremely traditional ones where you can hand-code it into your service.

I highly, highly advise at least a basic monitoring/alerting tool to at least monitor whether your services are alive and running properly (and of course I advocate running `dehydrated` from a system service, even if it's cron firing a sysvinit service, though a service/timer pair with systemd is what the cool kids should be doing). If you have your monitoring email you every day or few to let you know all is well, then you know the mail server and/or monitoring isn't down. :)

And all this is stopping short of the discussion of making a bad problem worse (50x due to capacity on the remote server) or hitting API limits. LE is a public service and we should all treat it responsibly.

lukas2511:
Retry logic may at some point be implemented, but it will require major changes to most of the loop behavior of dehydrated. This simple implementation of just looping again and again until it maybe succeeds at some point seems hacky to me. It's an acceptable solution for users with lots of network issues, but not something I want to upstream right now.

Issue #385, opened 2017-04-30
Support downloading OCSP responses


evaryont:
This is to work around an nginx bug, but perhaps it'd be useful for others, too.

It's be nice to be able to tell dehydrated to also download the OCSP response and keep it up-to-date as part of `--cron`. That way I can configure nginx (with [`ssl_stapling_file`](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_stapling_file)) to read the response at start. Right now, at least with nginx 1.12.0 on CentOS 7, when nginx starts it does not immediately request the OCSP and so doesn't staple it to the response. With OCSP_MUST_STAPLE enabled, Firefox will complain when you first access the server, and for a few more seconds beyond that too. Saving the response to disk should in theory prevent this problem. Perhaps `//etc/dehydrated/certs/domain.name/ocsp.der`?

txr13:
I believe that Let's Encrypt only signs the OCSP responses for a few days... Depending on how often a user runs their cron job, the response may expire.

I'd be inclined to write something outside of dehydrated to handle that, to avoid any instances of OCSP responses expiring on folks in between cron jobs.

evaryont:
For Let's Encrypt, the OCSP responses are valid for 1 week. So users would need to run the cron job at least once every 7 days, probably a little bit more often than that to allow them time to react to failures. Reading up on nginx's behaviour, it seems like if the ocsp response isn't valid any more, it will go and query OCSP on it's own like normal anyways. So if the cron job is missed, it doesn't block the server.

On top of all of that, this is a niche situation. How many people have turned on OCSP Must Staple? And how many of them are using Nginx? I figure the numbers are small enough that it would *not* be enabled by default, and would be another configuration option. If it's enabled, manage the OCSP response.

Some references online:
- https://unmitigatedrisk.com/?p=241
- https://blog.crashed.org/nginx-stapling-busted/
- https://gist.github.com/sleevi/5efe9ef98961ecfb4da8

lukas2511:
I tried to implement this in ee75c5dca7ea6a9f2500137683077a89fb444717, not sure how well this works, feel fry to play around with it. Set `OCSP_FETCH="yes"` in config to enable, it should write `ocsp-$timestamp.der` files and link the newest version to `ocsp.der`.

it-can:
I enabled this option and regenerated, but no ocsp file is showing up...?

lukas2511:
[...]

PR #457, opened 2017-12-08
be more verbose on OCSP stapling file updates

a-tze:
related to issue #385 
The OCSP update basically works like a charm, but it was making me pretty nervous to not see anything in the output related to OCSP when configuring it.
This is because the only echo was executed after the age check said an update is needed, so I propose to change that.


lukas2511:
Thanks for the suggestion, I think your changes were a bit too verbose, but I added a notice for skipped updates.

Issue #462, opened 2018-01-08
Give domains.txt file as cli param


rudiedirkx:
Is it possible to give a different `domains.txt` with the command? I know I can give a different config, and change `DOMAINS_TXT` there, but as a direct param would be nice, since the rest of the config doesn't change.

rudiedirkx:
Discussion in PR #463

PR #463, opened 2018-01-08
--domain param accepts filename to change domains.txt

rudiedirkx:
As stated in #462, I have several `domains.txt`s. This change makes different configs with different `DOMAINS_TXT`s unnecessary.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -229,6 +229,7 @@ load_config() {
   [[ -n "${PARAM_NO_LOCK:-}" ]] && LOCKFILE=""
 
   [[ -n "${PARAM_HOOK:-}" ]] && HOOK="${PARAM_HOOK}"
+  [[ -n "${PARAM_DOMAINS_TXT:-}" ]] && DOMAINS_TXT="${PARAM_DOMAINS_TXT}"
   [[ -n "${PARAM_CERTDIR:-}" ]] && CERTDIR="${PARAM_CERTDIR}"
   [[ -n "${PARAM_CHALLENGETYPE:-}" ]] && CHALLENGETYPE="${PARAM_CHALLENGETYPE}"
   [[ -n "${PARAM_KEY_ALGO:-}" ]] && KEY_ALGO="${PARAM_KEY_ALGO}"
@@ -1379,7 +1380,9 @@ main() {
       --domain|-d)
         shift 1
         check_parameters "${1:-}"
-        if [[ -z "${PARAM_DOMAIN:-}" ]]; then
+        if [[ -f "${1}" ]]; then
+          PARAM_DOMAINS_TXT="${1}"
+        elif [[ -z "${PARAM_DOMAIN:-}" ]]; then
           PARAM_DOMAIN="${1}"
         else
           PARAM_DOMAIN="${PARAM_DOMAIN} ${1}"


txr13:
Why would you not just point to a different config file (using `-f`) that references a different domains file? Yes, a new parameter makes that unnecessary, but I don't understand the need for it.

If you have several different domain lists, you're probably using different parameters for them in your config files anyway. If they all used the same parameters, you'd be using a single, unified list of domains.

rudiedirkx:
Several lists. Same exact config. Is that strange?

I don't override much config, but I don't want to repeat 4 vars for every list. If HOOK is a param, DOMAINS_TXT should definitely be a param IMO.

txr13:
That’s strange, at least to me. If you needed a per-certificate override, you could use the per-certificate config options. Otherwise, I’d expect one list of domains per general set of config options.

I guess what I’m not understanding is why you have several different domain lists for one set of config options.

alainwolf:
I have a small amount of certificates with special options (i.e. OCSP_MUST_STAPLE breaks mail client connections) so its a "no" there for mail servers. I use per-cert directory configs for those. But then I have two separate large groups of certs, the ones who can have EC certs and the ones who have to stick to RSA for compatibility with legacy clients (i.e. Android 7.0). I ended up creating both RSA and EC certs for all my servers and then choose the desired one in the server config. And I am fine with that, but wanted to point out a possible scenario of Rudies request.

rudiedirkx:
> why you have several different domain lists for one set of config options

The rest of the config options are also params. In this case `--challenge` and `--hook`.

Why some config options as cli params, and not others? I just missed this one, so I added it.

lukas2511:
@rudiedirkx They are params because they are useful (or even required) for controlling dehydrated by configuration management software like ansible, puppet, etc. They are working but are (mostly) not really designed to be used in combination with the "cron" functionality.

If you don't want to set all config variables you could easily source a base config in some sub-configs, and only override the DOMAINS_TXT option.

Or just concatenate your lists if nothing is really different? If everything else stays the same there really is no need to keep separate files, and this would be really just a single command to execute before dehydrated itself is running. You could even run this command inside your config, because that's just a shell script anyway.

If you really need separate files and don't want to modify your config or concatenate the lists or whatever else, you could also do something like `DOMAINS_TXT="${DEHYDRATED_DOMAINS_TXT:-}"` and use the environment variable to control which file to use.:`DEHYDRATED_DOMAINS_TXT=/etc/dehydrated/the-other-domains.txt dehydrated -c`.

Or do something like `dehydrated -f <(echo "DOMAINS_TXT=/etc/dehydrated/the-other-domains.txt"; cat /etc/dehydrated/config) -c`, or use a tmpfile, building a wrapper to do this is absolutely trivial.

For me this parameter really seems extremely unnecessary since the goal can already by achieved by a ton of other means and every reason I could think of for having two files would require separate configs to begin with.

Also your suggested implementation isn't that great, just think back when dehydrated was called letsencrypt.sh, which is a valid domain... every time I would have tried to refresh the certificate of that domain using `-d letsencrypt.sh` it would have tried parsing the script itself as domains list 💣 

So yea... Sorry but this won't be merged and I don't want this functionality.

Something that might come close would be a parameter to override single config variables, like `dehydrated --var DOMAINS_TXT=/etc/dehydrated/the-other-domains.txt --cron`, if that would be cleanly implemented I'd welcome it.

rudiedirkx:
Okay okay okay, I'm convinced. And honored to receive such a lengthy response! Separate configs it is.

PR #466, opened 2018-01-19
Fix grammer error in the manpage

mapreri:
"allows to" requires a subject (e.g. "allows one to"), without it's just
syntactically wrong.  Change the verb entirely to workaround the
problem.


jangrewe:
*grammar ;-)

PR #467, opened 2018-01-20
openssl: do not use -out when generating private keys

intelfx:
`openssl genrsa` and `openssl ecparam -genkey` tend to chmod the output
file to make the private key unreadable by group and others. This breaks
a workflow where admin uses POSIX ACLs to manage keys' access rights,
because POSIX ACLs reuse group permissions for the "ACL mask", hence
forcing group permissions to 0 with plain chmod() masks away all
extended ACLs.

By redirecting genrsa output from stdout we avoid this behavior while
staying secure even if ACLs are not used because we explicitly set
`umask 077` in the beginning of the script.

Example of the broken behavior:
```
# pwd                 
/etc/admin/certs/intelfx.name

# getfacl .
# file: .
# owner: letsencrypt
# group: letsencrypt
user::rwx
group::---
group:http-cert:r-x
mask::r-x
other::---
default:user::rwx
default:group::---
default:group:http-cert:r-x
default:mask::r-x
default:other::---

# getfacl cert.pem
# file: cert.pem
# owner: letsencrypt
# group: letsencrypt
user::rw-
group::---
group:http-cert:r--
mask::r--
other::---

# getfacl privkey.pem           
# file: privkey.pem
# owner: letsencrypt
# group: letsencrypt
user::rw-
group::---
group:http-cert:r--             #effective:---
mask::---
other::---
```

Here I run dehydrated as `letsencrypt:letsencrypt` and use group `http-cert` for all applications that need access to the certificates and the private key. The output directory is modified as follows:
```
setfacl -m u::rwX,g::0,o::0 -m g:http-cert:rX .
setfacl -d -m u::rwX,g::0,o::0 -m g:http-cert:rX .
```

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -309,7 +309,7 @@ init_system() {
       fi
 
       echo "+ Generating account key..."
-      _openssl genrsa -out "${ACCOUNT_KEY}" "${KEYSIZE}"
+      _openssl_stdout genrsa "${KEYSIZE}" > "${ACCOUNT_KEY}"
       register_new_key="yes"
     fi
   fi
@@ -426,6 +426,22 @@ rm_json_arrays() {
 _openssl() {
   set +e
   out="$("${OPENSSL}" "${@}" 2>&1)"
+  res=$?
+  set -e
+  if [[ ${res} -ne 0 ]]; then
+    echo "  + ERROR: failed to run $* (Exitcode: ${res})" >&2
+    echo >&2
+    echo "Details:" >&2
+    echo "${out}" >&2
+    echo >&2
+    exit ${res}
+  fi
+}
+
+# Same, but when we need OpenSSL's stdout intact
+_openssl_stdout() {
+  set +e
+  { out="$("${OPENSSL}" "${@}" 2>&1 1>&3-)"; } 3>&1
   res=$?
   set -e
   if [[ ${res} -ne 0 ]]; then
@@ -875,8 +891,8 @@ sign_domain() {
     echo " + Generating private key..."
     privkey="privkey-${timestamp}.pem"
     case "${KEY_ALGO}" in
-      rsa) _openssl genrsa -out "${certdir}/privkey-${timestamp}.pem" "${KEYSIZE}";;
-      prime256v1|secp384r1) _openssl ecparam -genkey -name "${KEY_ALGO}" -out "${certdir}/privkey-${timestamp}.pem";;
+      rsa) _openssl_stdout genrsa "${KEYSIZE}" > "${certdir}/privkey-${timestamp}.pem";;
+      prime256v1|secp384r1) _openssl_stdout ecparam -genkey -name "${KEY_ALGO}" > "${certdir}/privkey-${timestamp}.pem";;
     esac
   fi
   # move rolloverkey into position (if any)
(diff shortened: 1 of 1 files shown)


intelfx:
Note that I had to non-trivially change the definition of `_openssl()` to let stdout through —  while my bash-fu is mostly fine, I don't really know if this is OK semantically.

intelfx:
Ugh, so it is _not_, at least when checking certificate for corruption it makes certificate details leak to the terminal. OK, made it a separate wrapper which we use only for genrsa/genkey.

lukas2511:
Hey, which version of openssl were you using? I just tried this with 1.1.0(f,g) and the output had the correct acl set. I'd rather not implement any more wrappers if this just affects outdated openssl versions anyway...

intelfx:
```
$ export LC_ALL=C

$ pacman -Qi openssl | grep Version
Version         : 1.1.0.g-1

$ getfacl .
# file: .
# owner: operator
# group: users
user::rwx
group::---
group:http-cert:r-x
mask::r-x
other::---
default:user::rwx
default:group::---
default:group:http-cert:r-x
default:mask::r-x
default:other::---

$ openssl genrsa -out key1.rsa 4096
Generating RSA private key, 4096 bit long modulus
...................++
..............++
e is 65537 (0x010001)

$ openssl genrsa 4096 > key2.rsa
Generating RSA private key, 4096 bit long modulus
..........................................................................................................++
.......................................................................................++
e is 65537 (0x010001)

$ getfacl key1.rsa key2.rsa
# file: key1.rsa
# owner: operator
# group: users
user::rw-
group::---
group:http-cert:r-x             #effective:---
mask::---
other::---

# file: key2.rsa
# owner: operator
# group: users
user::rw-
group::---
group:http-cert:r-x             #effective:r--
mask::r--
other::---
```

lukas2511:
Ok sorry you are right, something weird is going on with the file permissions. I'll have a look at your patch in the next days and it will probably make it into the 0.6.1 release.

intelfx:
@lukas2511, any progress? This is literally a no-brainer.

lukas2511:
@intelfx sorry, lost track of this, can you verify that the patch below is working fine with acls? I didn't like the wrapper so I'm working with temporary files, should do the trick...

```
diff --git a/dehydrated b/dehydrated
index 2738495..80014e2 100755
--- a/dehydrated
+++ b/dehydrated
@@ -340,7 +340,10 @@ init_system() {
       fi
 
       echo "+ Generating account key..."
-      _openssl genrsa -out "${ACCOUNT_KEY}" "${KEYSIZE}"
+      local tmp_account_key="$(_mktemp)"
+      _openssl genrsa -out "${tmp_account_key}" "${KEYSIZE}"
+      cat "${tmp_account_key}" > "${ACCOUNT_KEY}"
+      rm "${tmp_account_key}"
       register_new_key="yes"
     fi
   fi
@@ -929,10 +932,13 @@ sign_domain() {
     if [[ ! -r "${certdir}/privkey.pem" ]] || [[ "${PRIVATE_KEY_RENEW}" = "yes" ]]; then
       echo " + Generating private key..."
       privkey="privkey-${timestamp}.pem"
+      local tmp_privkey="$(_mktemp)"
       case "${KEY_ALGO}" in
-        rsa) _openssl genrsa -out "${certdir}/privkey-${timestamp}.pem" "${KEYSIZE}";;
-        prime256v1|secp384r1) _openssl ecparam -genkey -name "${KEY_ALGO}" -out "${certdir}/privkey-${timestamp}.pem";;
+        rsa) _openssl genrsa -out "${tmp_privkey}" "${KEYSIZE}";;
+        prime256v1|secp384r1) _openssl ecparam -genkey -name "${KEY_ALGO}" -out "${tmp_privkey}";;
       esac
+      cat "${tmp_privkey}" > "${certdir}/privkey-${timestamp}.pem"
+      rm "${tmp_privkey}"
     fi
     # move rolloverkey into position (if any)
     if [[ -r "${certdir}/privkey.pem" && -r "${certdir}/privkey.roll.pem" && "${PRIVATE_KEY_RENEW}" = "yes" && "${PRIVATE_KEY_ROLLOVER}" = "yes" ]]; then
```

lukas2511:
I tested it myself and found one more bug (chain files were missing the attributes), it's now commited with 52c2c199945bd59a6f659f2aaff4c06bc6959c94. If you find anything else that's missing the attributes please let me know.

intelfx:
@lukas2511 Yep, works for me. I find the solution via temporary files to be dirtier, but well, whatever.

lukas2511:
@intelfx well, functionality wise i think your approach makes more sense, mine has really unnecessary tempfiles, but I think it's more readable and it duplicates less functionality. I'll eventually work on this in the future, but for now this does the job.

PR #469, opened 2018-01-28
Add ^~ to nginx location block

pandark:
To make sure it is not overridden.
> http://nginx.org/en/docs/http/ngx_http_core_module.html#location :
> If the longest matching prefix location has the “^~” modifier then regular expressions are not checked.


lukas2511:
Merged, thank you.

PR #472, opened 2018-02-03
Add example hook script for Gitlab pages

jspricke:


lukas2511:
Nice hook script, but please add it to the wiki. I'm keeping examples inside the repository at a bare minimum for now.

PR #473, opened 2018-02-03
Fix globbing of CONFIG_D *.sh files

GUI:
With the globbing changes made in 61083cf52231e09e3a9b599d602c06978704ec57 to globally disable globbing by default, this broke the ability to load the CONFIG_D `*.sh` files.

This re-enables globbing when reading these `*.sh` files and then disables it again afterwards. Note that this also keeps globbing enabled inside the loop, when sourcing the individual `*.sh` files for backwards compatibility (so if the individual config scripts relied on the default of globbing being enabled, there won't be any change in behavior).

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -158,6 +158,9 @@ load_config() {
       _exiterr "The path ${CONFIG_D} specified for CONFIG_D does not point to a directory."
     fi
 
+    # Allow globbing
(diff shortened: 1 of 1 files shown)


lukas2511:
Ooops... missed that one... Thanks!

PR #485, opened 2018-02-22
fixed "sed: invalid option -- 'E'"

edieterich:
While hitting a Let's Encrypt rate limit dehydrated printed this error:

``sed: invalid option -- 'E'``

This is caused by one occurrence in dehydrated where sed is used directly instead of the _sed wrapper. This pull requests fixes this.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -713,7 +713,7 @@ sign_csr() {
     fi
 
     # Check if authorization has already been validated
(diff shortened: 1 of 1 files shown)


lukas2511:
Thx, merged in a6a07779ad54f06fd75826bc357fafec2047cea2

PR #502, opened 2018-03-14
Improve documentation on wildcards

JohnVillalovos:
Improve the documentation on how to use wildcard domains.

Change-Id: Ie2ea0b3bb397d3a78113b2c3d9280e418a229d95

### DIFF
--- a/docs/domains_txt.md
+++ b/docs/domains_txt.md
@@ -1,22 +1,72 @@
-### domains.txt
+## domains.txt
 
-dehydrated uses the file `domains.txt` as configuration for which certificates should be requested.
+dehydrated uses the file `domains.txt` as configuration for which certificates
+should be requested.
 
 The file should have the following format:
 
 ```text
+example.org
 example.com www.example.com
 example.net www.example.net wiki.example.net
+```
+
+This states that there are the following certificates:
+  * `example.org` without any *alternative names*
+  * `example.com` with an *alternative name* of `www.example.com`
+  * `example.net` with the *alternative names*: `www.example.net` and
+    `wiki.example.net`
+
+### Aliases
+
+You can define an *alias* for your certificate which will (instead of the
+primary domain) be used as the directory name under your `CERTDIR` and for a
+per-certificate lookup. This is done using the `>` character.  This allows
+multiple certificates with identical sets of domains but different
+configuration to exist.
+
+Here is an example of using an *alias* called `certalias` for creating the
+certificate for `example.net` with *alternative names* `www.example.net` and
+`wiki.example.net`. The certificate will be stored in the directory `certalias`
+under your `CERTDIR`.
+
+```text
 example.net www.example.net wiki.example.net > certalias
 ```
 
-This states that there should be two certificates `example.com` and `example.net`,
-with the other domains in the corresponding line being their alternative names.
+### Wildcards
 
-You can define an alias for your certificate which will (instead of the primary domain) be
-used as directory name under your certdir and for a per-certificate lookup.
-This allows multiple certificates with identical sets of domains but different configuration
-to exist.
+Support for wildcards was added by the ACME v2 protocol.
 
-Certificates with a wildcard domain as first (or only) name require an alias to be set.
-Aliases can't start with `*.`.
+Certificates with a wildcard domain as the first (or only) name require an
+*alias* to be set.  *Aliases* can't start with `*.`.
+
+For example to create the wildcard for `*.service.example.com` your
+`domains.txt` could use the *alias* method like this:
+
+```text
+*.service.example.com > star_service_example_com
+```
+
+This creates a wildcard certificate for only `*.service.example.com` and will
+store it in the directory `star_service_example_com` under your `CERTDIR`. As a
+note this certificate will **NOT** be valid for `service.example.com` but only
+for `*.service.example.com`. So it would, for example, be valid for
+`foo.service.example.com`.
+
+
+Another way to create it is using *alternative names*. For example your
+`domains.txt` could do this:
+
+```text
+service.example.com *.service.example.com
+eggs.example.com *.ham.example.com
+```
+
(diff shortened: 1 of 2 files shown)


jangrewe:
The new CA URL is already the default, so the first part of your PR would probably confuse some people.
The domains.txt change is a nice touch, though. Kinda obvious, but probably still a good idea. ;-)

txr13:
I’m somewhat concerned that the update would be confusing. We have the note about needing to use an alias if the wildcard is the first or only domain... followed by a “for example” that doesn’t actually provide an example of that behavior at all.

The example given is good—for a certificate which has the wildcard as not the first or only domain. Set as an “example” (implicitly) of the alias functionality, it could just be more confusing.

JohnVillalovos:
> I’m somewhat concerned that the update would be confusing. We have the note about needing to use an alias if the wildcard is the first or only domain... followed by a “for example” that doesn’t actually provide an example of that behavior at all.

Can you give an example of what you think it should be? That was the only way I could figure out how to make a wildcard work.

JohnVillalovos:
> Kinda obvious, but probably still a good idea. ;-)

Well since it took me about 10 minutes trying to figure it out, it didn't seem obvious to me :)

I first tried.

    *.service.example.com

Then:

    *.service.example.com service.example.com

Then started adding debug code before I figured it out.

jangrewe:
uhm, but it states that so right at the end of the document... anyways, i agree that this part of the change doesn't hurt, while the first part probably would.

JohnVillalovos:
> uhm, but it states that so right at the end of the document... anyways, i agree that this part of the change doesn't hurt, while the first part probably would.

That is my point. It didn't make sense to me. In one part of the document it talks about ``alternative names`` in another part it talks about ``alias``. I'm not sure if they are the same thing or not.

> Certificates with a wildcard domain as first (or only) name require an alias to be set.
>  Aliases can't start with `*.`.

To me that implied that the wildcard domain is first, since it uses the word "first".

JohnVillalovos:
> The new CA URL is already the default, so the first part of your PR would probably confuse some people.

Okay. One question though. If someone has already been using dehydrated do they need to change the CA, or is it already changed? When I changed it, then I had to register and accept the terms. So it made me think it was not using the v1 endpoint.

txr13:
> Can you give an example of what you think it should be? That was the only way I could figure out how to make a wildcard work.

For example, you can use
`service.example.com *.service.example.com`
to create a certificate which covers both the domain itself and all subdomains under it. (Which is an extremely common use case!) The certificate directory name will be "service.example.com".

If you wish to create a certificate which _starts with_ a wildcard domain, you must use an alias. Your domains.txt file would then look like this:
`*.service.example.com > certificate-alias`
In this case, the certificate directory name will be "certificate-alias"

The reason why this is required is because the first domain of a certificate is usually used as the directory name. However, if the first domain on a certificate is a wildcard, that would result in a directory name of `*.service.example.com`. As the star character is also used in many filesystems to represent a wildcard, it can lead to confusion or unexpected/unusual behavior while trying to navigate the filesystem. For this reason, if you wish to use a wildcard domain as the first domain in a certificate, an alias must be set so that the directory in the filesystem can be used without potential issue.

txr13:
> If someone has already been using dehydrated do they need to change the CA, or is it already changed?

If the CA is set to Let's Encrypt's V2 endpoint (whether manually or by default), _and_ an account exists for the V1 endpoint, `dehydrated` will symlink the V1 account to a V2 account transparently.

JohnVillalovos:
@txr13 Thanks! That cleared up things a lot for me. I will make some updates :)

lukas2511:
I took a quick glance at your changes, looks good and I'll probably merge it later today or tomorrow.

Better documentation is something I'm going to be working on a lot more in the near(ish) future :wink:

txr13:
@JohnVillalovos I like the new changes! I think that's much clearer. There's one more thing I want to mention. It's not so much a documentation issue, so much as an additional point that _may_ influence the documentation, or merit additional explanation.

The catch is this: a certificate for `*.service.example.com` is **not valid** for `service.example.com`. That wildcard is only valid for the subdomains _under_ service.example.com. Which is why it's so common to see a certificate which has `service.example.com` and then `*.service.example.com` as an alternate name.

That's why `dehydrated` requires an alias, rather than simply stripping off the wildcard character. (Which was a potential solution discussed at the time.) This is not to mention the fact that, if you have a certificate for `service.example.com` as well, now the two certificates will collide.

My _suggestion_ would be to give an example with a proper alias, rather than simply stripping off the wildcard character. You may even wish to include the note about the fact that a wildcard doesn't cover the parent domain--I know that caught me by surprise the first time I ran into it. Because my guess is, the vast majority of users are going to want something that covers `service.domain.com` **and** everything underneath it.

JohnVillalovos:
> The catch is this: a certificate for *.service.example.com is not valid for service.example.com. That wildcard is only valid for the subdomains under service.example.com. Which is why it's so common to see a certificate which has service.example.com and then *.service.example.com as an alternate name. (In which case, no alias is needed at all.)

I'll add a note that about ``*.service.example.com`` is not valid for ``service.example.com``

> My suggestion would be to give an example with a proper alias, rather than simply stripping off the wildcard character. You may even wish to include the note about the fact that a wildcard doesn't cover the parent domain--I know that caught me by surprise the first time I ran into it. Because my guess is, the vast majority of users are going to want something that covers service.example.com and everything underneath it.

Is the example at the bottom using an alternative name not good? I can make a note that it supports both ``*.service.example.com`` and ``service.example.com``

I'll push up a new patch.

JohnVillalovos:
> My suggestion would be to give an example with a proper alias, rather than simply stripping off the wildcard character.

I didn't quite understand this at first, but I think I do now. I updated the patch to take that into account.

lukas2511:
Thanks, I merged your changes in 7c40c727a0e933382d72fa689ff3bea294895326 :+1:

PR #512, opened 2018-03-17
add example for wildcard entry

shmick:


lukas2511:
Hey, there already is a pullrequest open in #502 which adds an example and improves the documentation. Thanks anyway!

PR #538, opened 2018-04-14
deploy_ocsp timestamp parameter in example hook file is wrong

EmTeedee:
probably copy-paste from further up, timestamp is parameter 3


lukas2511:
Whoopsie. Thanks! Merged in 53c458c318c04a287eb9e786a29e8d056e88e432.

PR #553, opened 2018-04-30
Move hook challenge deployment into loop

jandryuk:
Certain DNS services (Duck DNS) only allow a single TXT record for a
domain and all subdomains.  So the last TXT value will overwrite all
early values.  To support such a service, setting TXT records
must be done in the challenge loop.  That way the TXT record is set and
immediately checked before going on to the next one.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -766,22 +766,19 @@ sign_csr() {
 
   # Deploy challenge tokens
   if [[ ${num_pending_challenges} -ne 0 ]]; then
-    echo " + Deploying challenge tokens..."
     if [[ -n "${HOOK}" ]] && [[ "${HOOK_CHAIN}" = "yes" ]]; then
+      echo " + Deploying challenge token chain..."
       "${HOOK}" "deploy_challenge" ${deploy_args[@]}
-    elif [[ -n "${HOOK}" ]]; then
-      # Run hook script to deploy the challenge token
-      local idx=0
-      while [ ${idx} -lt ${num_pending_challenges} ]; do
-        "${HOOK}" "deploy_challenge" ${deploy_args[${idx}]}
-        idx=$((idx+1))
-      done
     fi
   fi
 
   # Validate pending challenges
   local idx=0
   while [ ${idx} -lt ${num_pending_challenges} ]; do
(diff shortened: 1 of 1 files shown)


jobe1986:
Having just checked Duck DNS' documentation I would concur that it is the case that you can only set one TXT record at a time.

For those interested the API in question is documented at https://www.duckdns.org/spec.jsp

txr13:
@lukas2511 I'm curious about this from a purely technical / theoretical standpoint. Would something like the following not be workable?

```
if hook_chain
  request all authorization tokens
  deploy all tokens
  validate all tokens
else
  for each SAN
    request authorization token
    deploy token
    validate token
  end for
end if
```

I would be hesitant to make that sort of change to the flow, as it needlessly burdens the CA with much more traffic. And really, a DNS provider that only allows a single TXT records for a domain and all subdomains is not standards-compliant, so I see no reason to double or triple the CA's burden for the sake of supporting a non-compliant DNS provider. Again, I'm just interested in the theory here.

lukas2511:
@txr13 sorry it was late at night and I was confused and a bit annoyed, the real problem is dns caching on CA side. If I deploy one challenge after another the CA keeps the old record up to 300 seconds in cache (min. depending on your own dns ttl), so if somebody would want to validate *.example.org and example.org (which both will be validated on example.org) this would fail basically everytime, in some cases maybe even if your ttl would be set to 1. I was thinking about changing the behaviour depending on "duplicate" authorization names, but I'd rather avoid it, and up to now basically everybody who screamed "change it back to the old way!1" just had to modify their hook script to get it back working.

Duck DNS seems to me like a very odd provider, but I guess if somebody would contact them and explain them the situation they'd probably just implement multiple txt records.

lukas2511:
I added a section about this to [the troubleshooting guide](https://github.com/lukas2511/dehydrated/blob/master/docs/troubleshooting.md#dns-invalid-challenge-since-dehydrated-060--why-are-dns-challenges-deployed-first-and-verified-later) and [created a ticket/thread](https://github.com/lukas2511/dehydrated/issues/554) for a more centralized place for feedback about bad DNS provider APIs.

PR #574, opened 2018-07-09
Adds keep_going parameter to example config file

RandieM:
It would be nice, if the keep_going parameter were listed in the example config file, otherwise it is pretty difficult to find.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -117,3 +117,6 @@
 
 # ACME API version (default: auto)
 #API=auto
+
+# Keep going after encountering an error while creating/renewing multiple certificates in cron mode
+#PARAM_KEEP_GOING="yes"


txr13:
This PR should not be accepted. The variable `PARAM_KEEP_GOING` is not read from the config file--there's no code to do so. Also, variables of the type `PARAM*` are not used for config-file settings. Those are used for command-line switches only.

As keep-going is set by a command-line switch (and only by a command-line switch), it is not appropriate or effective to add this to the example config file. The switch is well-documented already (in the README.md, in the man page, and in the built-in help readout).

RandieM:
That's funny, though, because putting this parameter in the config actually does the trick. Some feedback would be nice, instead of simply closing the request.

lukas2511:
@RandieM sorry, i was closing this after @txr13's feedback.

Yes the parameter will be read from config, like tons of other variables and even functions (that's part of what makes dehydrated so adaptable), but it's not an intended config parameter and I don't want it to be one.

I don't really like making skipping of errors configurable, if you are using it in a cron you can always just add the parameter so it doesn't disturb anything, but when run by hand I actually want issues like this to be seen and not accidentally be ignored.

RandieM:
OK, thanks a lot, @lukas2511, for your feedback; this was much more helpful in understanding why you do not wish to include this in the config.

Just one question though; could it harm in any way if somebody does choose to include it in their own config?
I am asking this, because from a hosting provider's point of view, it is much better to have one customer complaining about one website's SSL that did not get automatically renewed, than have 20 customers complaining for 20 different websites.

lukas2511:
@RandieM It's probably better to add the command line switch to a cronjob, that way when you run the script manually for debugging it will still halt on errors. Otherwise there shouldn't be any harm, but there is no guarantee that it will keep on working in the future.

RandieM:
OK, understood. Thanks a lot for all your clarifications. Feel free to close this.

Issue #575, opened 2018-07-11
DOMAINS_D not in example config


bllfr0g:
DOMAINS_D is not in the example config, but it is referenced in the doc for per-certificate configuration, so I think it is "officially supported" and should be there, no?

PR #582, opened 2018-07-28
issue #575

bllfr0g:
Documentation change to fix issue #575


lukas2511:
Thanks! I added this manually as I wanted to rephrase it slightly and this way was easier than merging the patch first.

PR #609, opened 2018-11-30
Ensure files are written to disk before symlink

frankvanbever:
We encountered an issue on a device using UBIFS where if a power cycle
happens close enough after invoking dehydrated only the metadata would
have been written and the symlink points to an empty file e.g.

-rw------- 1 root root  0 Nov 30 13:20 chain-1543580304.pem
lrwxrwxrwx 1 root root 20 Nov 30 13:20 chain.pem -> chain-1543580304.pem
-rw------- 1 root root  0 Nov 30 13:20 fullchain-1543580304.pem
lrwxrwxrwx 1 root root 24 Nov 30 13:20 fullchain.pem -> fullchain-1543580304.pem

Forcing a sync of the symlinked files prevents this from occurring.

Signed-off-by: Frank Vanbever <frank.vanbever@essensium.com>

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1049,6 +1049,9 @@ sign_domain() {
     rm "${tmpcert}" "${tmpchain}"
   fi
 
+  # Wait for hook script to sync the files before creating the symlinks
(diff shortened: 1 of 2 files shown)


frankvanbever:
Hi, sorry for the late reply. I've updated the patch to use the hook infrastructure. Could you take another look please?

frankvanbever:
Hi, sorry for the late reply. I've updated the patch to use the hook infrastructure. Could you take another look please?

PR #611, opened 2018-12-07
Defer verification of domain-specific config options for --cron

chrisnovakovic:
The `--cron` option supports the loading of domain-specific config files (via the `DOMAINS_D` option in the global config file), but all configuration options are verified before the domain-specific config file is loaded, meaning that an otherwise-valid configuration split across both the global and domain-specific config files raises an error. I first ran into this problem when I set a domain-specific `WELLKNOWN` path and no value for `WELLKNOWN` on a system where the default directory `/var/www/dehydrated` didn't exist, and the resulting behaviour felt counter-intuitive after I'd specifically requested a particular `WELLKNOWN` path for that domain:

```
# dehydrated --cron --keep-going
 + Creating chain cache directory /etc/dehydrated/chains
Processing domain.com with alternative names: a.domain.com b.domain.com 
 + Creating new directory /etc/dehydrated/certs/domain.com ...
ERROR: WELLKNOWN directory doesn't exist, please create /var/www/dehydrated and set appropriate permissions.
```

This PR defers verification of options that may be included in domain-specific config files until after any domain-specific config file has been read.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -93,15 +93,22 @@ hookscript_bricker_hook() {
 }
 
 # verify configuration values
+# Pass "nodomains" as argument #1 to skip verifying options that may appear in a
+# domain-specific configuration file
 verify_config() {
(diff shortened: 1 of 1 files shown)


lukas2511:
Sorry for the really really really late response... Closing this for now as I'm currently rewriting major parts of dehydrated, turning it inside-out (checking certificates and their specific config first, contacting CA later on), which makes this unnecessary.

PR #613, opened 2018-12-11
Add sleep timer for dns-01 challenges

Kage:
This adds an optional sleep timer when there are pending challenges and the challenge type is dns-01.  It allows time for DNS to propagate for some slower providers if needed.

To use it, enable `SLEEPTIME=X` in your config, where `X` is a positive integer value in seconds.  This will pause the script just after deploying the challenges (only if there were challenges) before continuing to the validation phase.  This works especially well when coupled with [lexicon](https://github.com/AnalogJ/lexicon) in a hook script.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@ This file contains a log of major changes in dehydrated
 
 ## Added
 - Initial support for tls-alpn-01 validation
+- Optional sleep timer when using dns-01 to allow DNS to propagate
 
 ## [0.6.2] - 2018-04-25
 ## Added
--- a/dehydrated
+++ b/dehydrated
@@ -784,6 +784,9 @@ sign_csr() {
         idx=$((idx+1))
       done
     fi
+
+    # Sleep if dns-01 and sleep time defined
+    [[ "${CHALLENGETYPE}" == "dns-01" ]] && [[ -n "${SLEEPTIME}" ]] && [[ ${SLEEPTIME} -gt 0 ]] && echo " + Sleeping for ${SLEEPTIME} seconds..." && sleep ${SLEEPTIME}
   fi
 
   # Validate pending challenges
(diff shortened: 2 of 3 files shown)


txr13:
Typically, I would expect the hook script to verify that all records have propagated before returning to dehydrated.

Kage:
That may not always be possible.  Lexicon, for example, does not provide any real means to do this, and adding a sleep timer in the deployment hook can create an exponential timeout due to being called once per domain being validated.  This PR offers a simple workaround that doesn't interrupt the overall flow of dehydrated and is backwards-compatible for configs that don't have the `SLEEPTIME` variable defined.

One alternative could also be adding a whole new hook function that's called where I currently have the Bash `sleep` defined.  This could be used to do a ready-check to verify if all challenges are finished and ready for validation.  Would you prefer that option instead?

txr13:
I don’t understand why it might not be possible for a hook script to verify the result is live and fully propagated to all authoritative servers. If the hook script as written by another package does not implement this, it should be modified to do so (or a wrapper hook written to fill in the missing functionality).

Where multiple records are being deployed at once (with HOOK_CHAIN=yes), it makes sense to deploy all records and then enter a testing loop to verify all records are propagated before returning. Where a single record must be deployed and verified at a time (due to DNS provider limitations), it makes even more sense to test propagation and return when ready, rather than hard-coding a fixed time delay for each cycle of deploy-wait-verify.

One example of this can be seen in the https://github.com/bennettp123/dehydrated-email-notify-hook script. The has_propagated function is called from within the deploy hook and does not return until every challenge is fully propagated to all authoritative servers for the relevant domain. No new hook function was required.

Kage:
Hmm.. Interesting approach, thanks for the info.  I suppose I had previously overlooked `HOOK_CHAIN`.  I'll try that out and see if it fits my use case, then close this issue if it does.

lukas2511:
As @txr13 suggested this should happen inside the hook-script, ideally doing actual checks. Adding a simple sleep is a quick'n'dirty solution, but especially with larger number of certificates this might actually slow things down quite a lot.

PR #616, opened 2018-12-15
export HOOK_CHAIN when calling HOOK

simondeziel:
A generic hook needs to know if HOOK_CHAIN is set to yes in order to
handle parameters/arguments accordingly.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -324,7 +324,7 @@ init_system() {
   fi
 
   # Export some environment variables to be used in hook script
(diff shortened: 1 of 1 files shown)


txr13:
This really shouldn't be necessary. Hook scripts should be coded in such a way that they can work without having to explicitly test for HOOK_CHAIN's value.

See https://github.com/bennettp123/dehydrated-email-notify-hook for one implementation of this.

simondeziel:
@txr13, you are absolutely right, thank you and sorry for the noise.

PR #617, opened 2018-12-19
support mksh as executing shell

mirabilos:
Basic support for running under [The MirBSD Korn Shell](http://www.mirbsd.org/mksh.htm), the Unix shell with probably the most users (as it’s the Android default shell, among others):

- set -f exists, nullglob doesn’t, but considering we use noglob, it shouldn’t be needed
- use shell extglobs instead of regular expressions


lukas2511:
As with your other changes this is something which isn't really my main goal. The script was written mainly for bash and zsh support was only added because the changes were minimal. Posix shell compliance or support for other shells (which then might need to get tested on various versions and systems..) isn't my goal. Again sorry for the late reply.

PR #618, opened 2018-12-19
require only basic RE of sed(1) and no POSIX character classes

mirabilos:
this is for portability but not as well-tested as my other changes, although it works for me (obviously) on MirBSD; the extra `#'` line is to repair my editor’s syntax highlighting after a crazy line

for simplicity, I based this off the code after PR #617 was applied, as it would otherwise have context conflicts (directly adjacent lines are changed by these two PRs), and I thought #617 pretty indisputable and trivial to apply anyway

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -45,7 +45,6 @@ _mktemp() {
 check_dependencies() {
   # just execute some dummy and/or version commands to see if required tools exist and are actually usable
   "${OPENSSL}" version > /dev/null 2>&1 || _exiterr "This script requires an openssl binary."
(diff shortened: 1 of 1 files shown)


lukas2511:
Supporting this gets more complicated with every bigger change, closing this as I'm not really considering this modification anymore. Sorry for the really really really late response.. and thanks anyway.

mirabilos:
Yes well, I’ll need to keep it because I need to run dehydrated on MirBSD, and as mksh author I might be not even considering installing GNU bash either…

PR #635, opened 2019-03-26
Add --single to only process a single line of domains.txt

azielke:
This adds the Parameter --single to only process a single certificate of domains.txt.

The value of --single is tried to match against the first domain of every line in domains.txt only.

With an domains.txt like this:
```
example.com www.example.com
example.net
```

An run with: dehydrated -c --single example.com
the certificate for "example.com www.example.com" will be generated.

dehydrated -c --single www.example.com would not generate any certificate.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -564,6 +564,7 @@ http_request() {
 
       # remove temporary domains.txt file if used
       [[ "${COMMAND:-}" = "sign_domains" && -n "${PARAM_DOMAIN:-}" && -n "${DOMAINS_TXT:-}" ]] && rm "${DOMAINS_TXT}"
+      [[ "${COMMAND:-}" = "sign_domains" && -n "${PARAM_SINGLE:-}" && -n "${DOMAINS_TXT:-}" ]] && rm "${DOMAINS_TXT}"
       exit 1
     fi
   fi
@@ -1177,6 +1178,10 @@ command_sign_domains() {
     mkdir "${CHAINCACHE}"
   fi
 
+  if [[ -n "${PARAM_DOMAIN:-}" && -n "${PARAM_SINGLE:-}" ]]; then
+    _exiterr "--single and --domain cannot be used at the same time"
+  fi
+
   if [[ -n "${PARAM_DOMAIN:-}" ]]; then
     DOMAINS_TXT="$(_mktemp)"
     if [[ -n "${PARAM_ALIAS:-}" ]]; then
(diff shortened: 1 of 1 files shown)


txr13:
What is the use case for this?

In normal use, we would want all certificates processed. For advanced use, we have the -d parameter.

I guess I’m wondering what makes this a better approach than parsing the domains.txt, locating the desired line, and using that to produce an appropriate -d parameter list.

jahir:
I gues the difference is that with `--single` you can select a whole line in `domains.txt`, i.e. the CN and all the SANs are included. With `--domain`, only a single domain name is processed.
I'm in favour for more elaborate options to specify domain names, the current domains.txt approach always seemed a little clumsy to me (and I scripted around it to mangle/replace domains.txt appropriately before running dehydrated). Not sure if `--single` is the way to go, but an improvement for sure nonetheless.

jobe1986:
You can specify --domain (-d) more then once, with the first being used as the CN, and additional being used as SANs.

azielke:
Yes, I agree that the use case might be limitted.

For me it is: I generate the domains.txt, but sometimes certificates get updated and a domain is added or removed. I want to run dehydrated for that certificate and want to know, if it could be obtained.

I had the Idea of just generating a temporary domains.txt and tell dehydrated to use that, but there is no parameter for it, so I would have to rewrite the config file.
Parsing the domains.txt would be possible aswell, but why parse the list outside (and maybe have to deal with certalias aswell), when all the functionality is in dehydrated. So this just seemed the easiest way for me.

Another use case might be renewing a single certificate with --force, instead of all certificates.

jobe1986:
> I had the Idea of just generating a temporary domains.txt and tell dehydrated to use that, but there is no parameter for it, so I would have to rewrite the config file.
You can specify a path to an alternative config file using the --config (-f) command line flag, so you would be able to do it that way without having to constantly edit your main config file.

azielke:
Yes, that would be possible. But for that you'd need to copy the line of domains.txt to a temp file, create another a tempfile, copy over the config, add/modify DOMAINS_TXT, run dehydrated and cleanup.

If everything is automated anyway, no problem - but if you want/need to do that manually, I like my way better ;)

lukas2511:
I'd like to keep features that are easily done in a single line of bash outside of the script, so I'm closing this pull request, sorry.

Something like this would fulfill your needs:

```
dehydrated -c -d "$(grep ^example.com domains.txt)"
```

lukas2511:
Also please keep in mind that dehydrated actually recognizes if a domain-set changed and will automatically refresh the certificate, so maybe you don't even need this feature?

PR #637, opened 2019-04-12
Default BASEDIR to /etc if installed to /usr/bin

polyzen:
This would simplify package installation with the following layout: https://paste.xinu.at/y55Udx/

If you're willing to accept this change, perhaps a more tailored warning should be used here.


polyzen:
Not sure what I was trying to accomplish with this.

PR #639, opened 2019-04-25
Wait if order status is processing

johanlundberg:
We noticed that our acme server implementation was not finished creating the certificate as fast as letsencrypt after order finalization. So this patch adds a wait for the order to get out of processing state.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -848,13 +848,26 @@ sign_csr() {
   fi
 
   # Finally request certificate from the acme-server and store it in cert-${timestamp}.pem and link from cert.pem
-  echo " + Requesting certificate..."
   csr64="$( <<<"${csr}" "${OPENSSL}" req -config "${OPENSSL_CNF}" -outform DER | urlbase64)"
   if [[ ${API} -eq 1 ]]; then
+    echo " + Requesting certificate..."
     crt64="$(signed_request "${CA_NEW_CERT}" '{"resource": "new-cert", "csr": "'"${csr64}"'"}' | "${OPENSSL}" base64 -e)"
     crt="$( printf -- '-----BEGIN CERTIFICATE-----\n%s\n-----END CERTIFICATE-----\n' "${crt64}" )"
   else
-    result="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' | clean_json | get_json_string_value certificate)"
+    echo " + Finalizing order..."
+    order="$(echo ${finalize} | sed 's,/*[^/]\+/*$,,')"
+    result="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' | clean_json)"
+    reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
+
+    while [[ "${reqstatus}" = "processing" ]]; do
+      echo " + Waiting for certificate processing..."
+      sleep 1
+      result="$(signed_request "${order}" '{}' | clean_json)"
+      reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
+    done
+
+    echo " + Requesting certificate..."
+    result="$(printf "${result}" | get_json_string_value certificate)"
     crt="$(signed_request "${result}" "")"
   fi
 


dehydrated
```diff
@@ -848,13 +848,26 @@ sign_csr() {
...
   else
-    result="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' | clean_json | get_json_string_value certificate)"
+    echo " + Finalizing order..."
+    order="$(echo ${finalize} | sed 's,/*[^/]\+/*$,,')"
+    result="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' | clean_json)"
+    reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
+
+    while [[ "${reqstatus}" = "processing" ]]; do
+      echo " + Waiting for certificate processing..."
+      sleep 1
```
Rogdham: One second may be too small? [The RFC](https://tools.ietf.org/html/rfc8555#section-8.2) hints server implementations not to retry faster than 5 to 10 seconds… so maybe do a `sleep 5` or `sleep 10`?

dehydrated
```diff
@@ -848,13 +848,26 @@ sign_csr() {
...
     crt64="$(signed_request "${CA_NEW_CERT}" '{"resource": "new-cert", "csr": "'"${csr64}"'"}' | "${OPENSSL}" base64 -e)"
     crt="$( printf -- '-----BEGIN CERTIFICATE-----\n%s\n-----END CERTIFICATE-----\n' "${crt64}" )"
   else
-    result="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' | clean_json | get_json_string_value certificate)"
+    echo " + Finalizing order..."
+    order="$(echo ${finalize} | sed 's,/*[^/]\+/*$,,')"
+    result="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' | clean_json)"
+    reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
+
+    while [[ "${reqstatus}" = "processing" ]]; do
```
Rogdham: I would say you should match `processing` as well.

If the status is `invalid`, what do you do here? just continue and the next request will fail (but good luck debugging)? Maybe test against that value as well?

Something like this maybe? (I haven't tested it)
```sh
while :
do
  case "${reqstatus}"
  in
    "processing" | "pending")
      echo "   > Order in status ${reqstatus}, retrying in a while..."
      sleep 5;
      ;;
    "valid")
      break;
      ;;
    *)
      _exiterr "Order in status ${reqstatus}"
      ;;
  esac
  result="$(signed_request "${order}" '{}' | clean_json)"
  reqstatus="$(printf '%s\n' "${result}" | get_json_string_value status)"
done
```

dehydrated
```diff
@@ -848,13 +848,26 @@ sign_csr() {
...
-  echo " + Requesting certificate..."
   csr64="$( <<<"${csr}" "${OPENSSL}" req -config "${OPENSSL_CNF}" -outform DER | urlbase64)"
   if [[ ${API} -eq 1 ]]; then
+    echo " + Requesting certificate..."
     crt64="$(signed_request "${CA_NEW_CERT}" '{"resource": "new-cert", "csr": "'"${csr64}"'"}' | "${OPENSSL}" base64 -e)"
     crt="$( printf -- '-----BEGIN CERTIFICATE-----\n%s\n-----END CERTIFICATE-----\n' "${crt64}" )"
   else
-    result="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' | clean_json | get_json_string_value certificate)"
+    echo " + Finalizing order..."
+    order="$(echo ${finalize} | sed 's,/*[^/]\+/*$,,')"
```
Rogdham: This is not always the right URL to get the certificate… I believe you need to get that from the `Location` header returned by the finalize request…

```sh
order="$(signed_request "${finalize}" '{"csr": "'"${csr64}"'"}' 4>&1 | grep -i ^Location: | awk '{print $2}' | tr -d '\r\n')"
```

Note that in that case you don't get the `result` variable to extract the `reqstatus`, maybe there is a simple way to get around that? Else, do the get `order` request as the first thing in the `while` loop in any case to retrieve the status.

johanlundberg:
Sadly I no longer have an environment to test these improvements but I am happy to make the changes you suggest or let you do which ever changes you see fit to this PR (or close it and fix it another way).

Rogdham:
No problem, I completely understand commitment issues in open source projects: sometimes you don't have time anymore, sometimes you just move on with your life or switch to an alternative software. In any case, your reply is really appreciated: now we know what to expect and can act accordingly. :+1: 

So I created #728 to solve the same issue taking into account what you did and the comments I posted in my review.

lukas2511:
Replaced by #728, thanks.

PR #656, opened 2019-07-08
libressl compability for ocsp_fetch

jnbr:
LibreSSL (2.9.2) requires the host header to be set the same way as OpenSSL <= 1.0 does.


lukas2511:
Is this valid for older libressl versions as well? Otherwise your patch would probably break compatibility with those.

jnbr:
Yes, I checked a few older releases and it's valid for them as well.
Since OpenSSL-1.1, the Host header is added by default, libressl never adopted this.

lukas2511:
This has been replaced by #784. Thanks anyway!

PR #663, opened 2019-08-19
Fix all ShellCheck errors and warnings

andersk:
It looks like someone did this before but the script has since regressed. Consider adding ShellCheck as a CI pass to make sure mistakes are caught proactively.


lukas2511:
replaced by e963438c5a5d58ce3f997eef1e8e25f3cb4238d8, thx anyway

Issue #230, opened 2016-06-27
Checking domain name of existing cert may fail


jlesage:
I have a scenario where, during server deployment, ${CERTDIR}/${domain}/cert.pem is pointing to a self-signed certificate.  This is done to make sure applications won't complain about missing certificate if  letsencrypt.sh didn't ran yet.

When running letsencrypt.sh in this scenario, the script fails because the existing certificate doesn't contains the string `DNS:`.

The problem is with the following line:

```
certnames="$(openssl x509 -in "${cert}" -text -noout | grep DNS: | _sed 's/DNS://g' | tr -d ' ' | tr ',' '\n' | sort -u | tr '\n' ' ' | _sed 's/ $//')"
```

I guess that not finding the `DNS:` string should not be considered as an error and should force a renew.

lukas2511:
Not going to fix this, if this is really necessary I suggest putting your certificate in fullchain.pem, as that is what you'd want to use later anyway and the script should work fine that way.

PR #666, opened 2019-08-26
Only check existing certs when necessary

stefaang:
Fixes https://github.com/lukas2511/dehydrated/issues/230

The expiry date check could be skipped too.

Some extra info:
I tried running dehydrated on an entware installation + lighttpd webserver based on [this](https://redmine.lighttpd.net/projects/lighttpd/wiki/HowToSimpleSSL#Lets-Encrypt-bootstrap-using-TLS-ALPN-01-verification-challenge-and-dehydrated-with-lighttpd-1453).
Unfortunately, the openssl package in entware doesn't support the -extensions req_ext flag, hence there's no DNS section in generated keys.
I tried running the script without any certificates, but lighttpd doesn't want to start when the certs in its config do not exist. Dehydrated doesn't want to generate certs when it fails to parse the DNS section in existing certs. A bit of a chicken/egg situation.

Hence, when forcing to generate certs anyway, do not parse the existing certs.. the main goal of that section is to enable the force flag anyway.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1304,7 +1304,7 @@ command_sign_domains() {
     fi
 
     # Check domain names of existing certificate
-    if [[ -e "${cert}" ]]; then
+    if [[ -e "${cert}" && "${force_renew}" = "no" ]]; then
       printf " + Checking domain name(s) of existing cert..."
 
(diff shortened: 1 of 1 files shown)


lukas2511:
merged, thx

PR #669, opened 2019-09-03
changing how we fetch the account id from the Location

aloopkin:
Relying on the account id being the last element in the URL could be unreliable... and is not working with some ACMEv2 implementations.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,12 +1,26 @@
 # Change Log
 This file contains a log of major changes in dehydrated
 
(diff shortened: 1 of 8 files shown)


Rogdham:
Thinking about this in the context of #727, what do we really need the account id for anyway? If we store the account url, we don't need the account id for anything than a debug line when updating the account… and we could show the account url there instead.

lukas2511:
Replaced by #727. Thanks!

Issue #674, opened 2019-09-16
dns-01: base-domain and wildcard: all challenge deploys are done before validation


fandauchytil:
### Problem

If "challenge type" is set to "dns-01" and in "domains.txt" are defined both fqdn and wildcard for that fqdn, script will do challenge for **both** of them **before** validation.

### Info

commit 05eda91a2fbaed1e13c733230238fc68475c535e

### Example 

(Below, I replaced my domain for 'example.com'.)

dehydrated.cfg
```
CHALLENGETYPE="dns-01"
DOMAINS_TXT="domains.txt"
``` 
domains.txt
```
example.com *.example.com
```
Output (full output is at the end):
```
# letsencrypt.sh' -f dehydrated.cfg' -c
...
 + Received 2 authorizations URLs from the CA
 + Deploying challenge tokens...
+ example.com TXT Z7ORPxfTRk-tfwjqRDetMmcMdno83uv9IPAVnsw3nJ8
+ example.com TXT 9yeYs0leymz9Xp9snRkPeZpc7pOEWoeQYpWeXqBBLrU  # <--- This rewrites the one above
... 
ERROR: Challenge is invalid! (returned: invalid) (result: {
    "detail": "Incorrect TXT record \"9yeYs0leymz9Xp9snRkPeZpc7pOEWoeQYpWeXqBBLrU\" found at _acme-challenge.example.com",
...
```
It sets up TXT for 'example.com' twice -- the first one is for domain 'example.com'  and the second one is for '*.example.com' (so the first one will be always rewritten). First challenge will always  fail, because it expects: `Z7ORPxfTRk..`, but it will get `9yeYs0leymz9...`.

### Full output from exmaple

```
Processing example.com with alternative names: *.example.com
 + Checking domain name(s) of egit commit -mxisting cert... unchanged.
 + Checking expire date of existing cert...
 + Valid till Oct 13 08:10:59 2019 GMT Certificate will expire
(Less than 30 days). Renewing!
 + Signing domains...
 + Generating private key...
 + Generating signing request...
 + Requesting new certificate order from CA...
 + Received 2 authorizations URLs from the CA
 + Handling authorization for example.com
 + Handling authorization for example.com
 + 2 pending challenge(s)
 + Deploying challenge tokens...
+ example.com TXT Z7ORPxfTRk-tfwjqRDetMmcMdno83uv9IPAVnsw3nJ8
+ Sleeping for 80s...
[...]

Issue #554, opened 2018-05-01
Discussion: Bad DNS provider APIs


lukas2511:
Some DNS providers only allow a single TXT record per domain. This can currently result in problems when trying to use domain sets like `example.org *.example.org` where two challenges are verified on the same domain.

Please read [tbd] before commenting on this thread!!!!!!!

If (and only if!) you are using one of those providers and there is no way to fix your hook-script, your DNS-provider doesn't want to change their API and you can't easily change to a different DNS provider please leave a comment on this thread containing your DNS provider and a link to the API documentation so I can verify the issue.

Please also check if somebody else already commented with that provider and in that case just give that post a thumbs-up or something.

wpyoga:
Hi, I have a workaround for this problem, with GoDaddy API. This provider does not let us add new TXT records with the same name, it always overwrites all existing records with the data from the latest PUT API call.

It's not perfect, because it creates a throwaway certificate in the process. So, without having to handle existing `_acme-challenge` records, we should have two lines in domains.txt:
```txt
*.example.com > example_com
*.example.com example.com > example_com
```

This works for simple personal servers, but will likely fail if the server has too many domains on it. What will happen is:
1. dehydrated will read the first line request a certificate with `CN=*.example.com`, and it will complete 1 single challenge
1. dehydrated will read the first line and checks the existing certificate, which has a different set of domains, so it will force a renew
1. dehydrated will request the new certificate, and it will receive 1 single challenge (because the domain `*.example.com` has been authorized recently)
1. certificate renewal will succeed

What do you guys think?

Notes: For GoDaddy, the proper solution requires a more advanced hook script. For some DNS providers like Duck DNS mentioned in #553 , this might be helpful... or not?

marcomsousa:
duckdns.org only allow a single TXT record per domain.

Since my scenario are:
- a.example.com CNAME  b.duckdns.org 
- _acme-challenge.a.example.com CNAME _acme-challenge.b.duckdns.org

Since 0.6.0 if I execute dehydrated with two domains:
`dehydrated --cron --hook ./hooks.sh --challenge dns-01 --domain a.example.com --domain b.duckdns.org`
Dehydrated try to deploy the two challenge at the same time. Only after, will responding the challenge for both domains. (parallel)

I can only validate if I can force dehydrated to do sequential validations.
Deploy 1 challenge, validate 1 challenge, clean, deploy 2 challenge, validate 2 challenge, clean. (sequential).

[...]

PR #675, opened 2019-09-16
FIX dns-01 domain + wildcard walidation (#674)

fandauchytil:
Fix for issue #674.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -790,19 +790,19 @@ sign_csr() {
     echo " + Deploying challenge tokens..."
     if [[ -n "${HOOK}" ]] && [[ "${HOOK_CHAIN}" = "yes" ]]; then
       "${HOOK}" "deploy_challenge" ${deploy_args[@]}
-    elif [[ -n "${HOOK}" ]]; then
-      # Run hook script to deploy the challenge token
-      local idx=0
-      while [ ${idx} -lt ${num_pending_challenges} ]; do
-        "${HOOK}" "deploy_challenge" ${deploy_args[${idx}]}
-        idx=$((idx+1))
-      done
     fi
   fi
 
   # Validate pending challenges
   local idx=0
   while [ ${idx} -lt ${num_pending_challenges} ]; do
+
+    if [[ -n "${HOOK}" ]] && [[ ! "${HOOK_CHAIN}" = "yes" ]]; then  # NOT HOOK_CHAIN (it's more reliable check than '$HOOK_CHAIN = no')
+      # Run hook script to deploy the challenge token
+      echo " + Deploying challenge tokens..."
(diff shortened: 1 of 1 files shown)


txr13:
How does this purport to fix the issue? The only changes are to move a non-chained hook’s deploy sequence to within the validation sequence. This will affect all hooks, not just those used for DNS.

The issue linked appears to be a case of one validation record being overwritten by another. That is a known issue with some DNS providers or hook scripts, but it is not a flaw in dehydrated. I would be very wary of trying to remedy it by introducing a change for all hook scripts regardless of whether they suffer this issue or not.

fandauchytil:
Yes, you are right. I didn't understand issue with "potential huge TTL" and its consequences. In my commit[1], I wanted to show what I need [1], and if I add an additional condition, it cloud be an option. But it looks like I'm anomaly (#554), so I have no problem with local patch. Thank you for your time.

[1] I did not found a better way on github how to show a prototype.
[2] I need (want) to do some sequential stuff.

fandauchytil:
I added variable 'HOOK_SEQ', it will trigger sequential "deploy challenge" for every "challenge verification".

I tested original functionality and it should work without breakage (code path is the same). Please look at it and tell me if it's ok, or if some more changes will be needed (or if you will not accept it by principle -- I'll have no problem with it, it's your project, but please let me know).

txr13:
So, having read through the code, I want to double-check my understanding of the proposed changes:

1. `HOOK_CHAIN` and `HOOK_SEQ` are mutually incompatible options, with `HOOK_CHAIN` overriding `HOOK_SEQ`.

2. No message will be logged regarding the cleaning of challenge tokens if `HOOK_CHAIN="no"` and `HOOK_SEQ="yes"`.

3. `HOOK_SEQ` is supported for use in a per-certificate config file.

Is my understanding correct? You may wish to add additional commits updating the appropriate documentation for per-certificate configs and the main config. You may also want to make it clear to the user that using `HOOK_CHAIN="yes"` will have the same effect as setting `HOOK_SEQ="no"` because they are mutually incompatible.

All in all, I still feel this is a very niche case to work around broken implementations which don't correctly support multiple TXT records per hostname. It would probably be better if people just moved to a working implementation instead. Still, that's not my decision to make. :)

fandauchytil:
Yes, you are correct in everything... I thought very hard about this and I'm not happy with this hack (I use it only on one legacy system I'm stuck with and after upgrade, there will be no real need for this). Dehydrated is excellent project and I don't want to taint it by this. Nonetheless, if you think, that I should finish it, I will.

Thank you very much for your time, I really appreciate it :).

lukas2511:
Closing this as it's not really relevant for most users and unless behaviour gets configurable this might cause problems for others.

PR #679, opened 2019-10-01
Add missing " in hook example

mookie-:
In the example sync_cert hook one `"` is missing. So it would be a broken hook script if you just uncomment that line.


lukas2511:

PR #693, opened 2019-12-01
docs: use `service` instead of `systemctl` in the examples

mapreri:
`systemctl` is specific to systemd, whereas `service` is a init system
agnostic command, that can run in all OSes I know about.

I've been asked to do it [in this Debian bug](https://bugs.debian.org/945908)

### DIFF
--- a/docs/examples/hook.sh
+++ b/docs/examples/hook.sh
@@ -87,7 +87,7 @@ deploy_cert() {
 
     # Simple example: Copy file to nginx config
     # cp "${KEYFILE}" "${FULLCHAINFILE}" /etc/nginx/ssl/; chown -R nginx: /etc/nginx/ssl
-    # systemctl reload nginx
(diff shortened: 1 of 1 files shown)


mckaygerhard:
well this commit are weak and incomplete amybe.. must depends on paths due some systems does not have "service" in their paths.. like debian 10 shit.. but devuan and debian 9 works perfectly

mapreri:
it's in `/usr/sbin` - something that makes a lot of sense since `service` is useful only for the system administrator and rarely is of any use for a regular user.  Therefore `root` can use `service` directly.

mckaygerhard:
@mapreri  seems you dont understan.. service are in /usr/sbin/ in your winbuntnu or wincentos.. but not in all systems.. that's the problem..

lukas2511:

PR #704, opened 2020-01-12
Get OS from /etc/os-version instead of /etc/issue

mkotich:


lukas2511:
Unfortunately `/etc/os-release` is only available on relatively new distributions, while `/etc/issue` is basically available everywhere. Can't merge it this way, but I guess I could add a change that it tries `os-release` first and `issue` as fallback... mhm...

mkotich:
The reason I suggested moving to os-release is it's a pretty solid new
standard, I tested that file (and the contents thereof) on cent7+,
ubuntu16+ and slackware14.

The problem with checking issues and falling back to os-release is issues
will exist in pretty much everything, but it's also a file full of nonsense
on cent and redhat
(here's cent7)
[matt@CentOS7 ~]# cat /etc/issue
\S
Kernel \r on an \m

It seems to work fine in Debian based systems though.

I don't have access to my slackware system from where I am right now, but I
can check when I get home...  As far as past use I believe /etc/issue is
mainly used to customize what shows up before a login prompt, so I don't
think this file is the MOST reliable path to discerning an OS.  The fact
that os-version will catch "most" OSs is the reason I added a "OS
Undetermined" if that file is not found.  I figure this way you get either
a "pretty name" like "Ubuntu 16.04.6 LTS" or "CentOS Linux 7 (Core)" or
"Slackware 14.2" or if that file does not exist for some reason, you get
"OS Undetermined" which prevents us from getting something ugly like in the
case of CentOS.  While it's true that os-version is generally new, I think
if you're casting your net to catch as many fish as possible the results
from this method might be cleanest?  That's my perspective at least...

Thanks!


-- 

Matt Kotich
Kotich.com

PR #716, opened 2020-03-16
Check for openssl/awk/sed and all are executable

### LINKED ISSUE #715: check_dependencies doesn't check exec status nor for awk/sed
`check_dependencies` uses `command -v` to determine the path (and presence) of
various binaries, it does not however check their execution status.

In the event `PATH` contains a matching name, but it is NOT executable,
`command -v` will still return it, so we must check for it's ability to be
executed.

Also, `awk` and `sed` aren't checked for which are used a fair bit.

krayon:
This fixes #715 by checking the commands are not just in path, but executable. It also checks for the presence of `awk` and `sed`.

**NOTE:** `[ -x "" ]` will fail correctly so if no file is found, the correct result will still be achieved.

Signed-off-by: Krayon <krayon.git@qdnx.org>

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -40,11 +40,14 @@ _mktemp() {
 # Check for script dependencies
 check_dependencies() {
   # just execute some dummy and/or version commands to see if required tools exist and are actually usable
+  [ -x "$(command -v "${OPENSSL}" 2>/dev/null)" ] || _exiterr "This script requires ${OPENSSL}."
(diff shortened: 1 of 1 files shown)


lukas2511:
Thanks for your input. I modified the dependency checks a bit and all your issues should be resolved with that. See a07c8d14f62c3b32e4f9e963dabc0691f55fcde6.

PR #718, opened 2020-03-16
Fix curl check regression and simplify

### LINKED ISSUE #717: curl check regression
Commit 0af7f388d6e63e74920e5f0cc9ae21f717d223d1 added an error code check for
older curl versions that return an initialisation error (2) on version check.
From my tests 7.10 from 2002 is the last to do this I believe FYI:

```diff
    check exitcode of `curl -V` because of some issue with ancient versions of curl returning error codes on checking the version number

+  # curl returns with an error code in some ancient versions so we have to catch that
+  set +e
+  curl -V > /dev/null 2>&1
+  set -e
+  retcode="$?"
+  if [[ ! "${retcode}" = "0" ]] && [[ ! "${retcode}" = "2" ]]; then
+    _exiterr "This script requires curl."
+  fi
```

A regression was introduced in 68274646bbb40e733d9fb5f3b0590d124a3aaea9 , that
breaks the `retcode` assignment as it's now coming from the `awk`, not the
`curl`:

```diff
    curl: use custom user agent (temporarily using a bit of 1337)

-  curl -V > /dev/null 2>&1
+  CURL_VERSION="$(curl -V 2>&1 | head -n1 | awk '{print $2}')"
```

krayon:
This fixes #717 by capturing the output into a string for later version extraction.

Additionally, it replaces the subshell calls to `head` and `awk` with basic
`bash` `read` which is 95+ times faster (in my tests, based on 100000
iterations), simpler and safer (no forking that could fail etc).

**NOTE:** You'll notice I'm not redirecting `stderr` anymore. Even the oldest
version of `curl` I could find (4.8 from 2000) still outputs to `stdout` so I'm
not sure why this is being captured.

Lastly, it also reuses the curl version information previously obtained for the
`command_version()` function.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -51,7 +51,7 @@ check_dependencies() {
   local _tmp="$(curl -V 2>/dev/null)"
   retcode="$?"
   set -e
-  CURL_VERSION="$(echo "${_tmp}" | head -n1 | awk '{print $2}')"
+  read _ CURL_VERSION _ <<<"${_tmp}"
   if [[ ! "${retcode}" = "0" ]] && [[ ! "${retcode}" = "2" ]]; then
     _exiterr "This script requires curl."
   fi
--- a/dehydrated
+++ b/dehydrated
@@ -48,9 +48,10 @@ check_dependencies() {
 
   # curl returns with an error code in some ancient versions so we have to catch that
   set +e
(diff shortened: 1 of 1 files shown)


lukas2511:
Thanks for your input. I modified the dependency checks a bit and all your issues should be resolved with that. See a07c8d14f62c3b32e4f9e963dabc0691f55fcde6.

PR #725, opened 2020-03-24
Fix typo "implicitly"

simon04:


lukas2511:

mckaygerhard:
@lukas2511  i make two articles.. for debian and devuan linux systems, but using Debian only ways. i mean using package guidelines, aybe some package guidelines must be take in consideratin into the new documentation.

please keep me updated to also make alpine related documentation.

PR #727, opened 2020-03-30
Fix account URL

### LINKED ISSUE #671: dehydrated is not using the canonical Account URL
When you `dehydrated --register` a new ACMEv2 account, dehydrated correctly extracts the Account URL from the `Location` header in the registration response:

https://github.com/lukas2511/dehydrated/blob/018254974cb6cbf93ba9249b06b2d923e37a9e88/dehydrated#L414

However, on subsequent request (e.g. in `dehydrated -c`), dehydrated does not use that `ACCOUNT_URL`, and instead assumes that the account URL is constructed from the `newAccount` resource in the directory:

https://github.com/lukas2511/dehydrated/blob/018254974cb6cbf93ba9249b06b2d923e37a9e88/dehydrated#L421

In the case when you run against [Pebble](https://github.com/letsencrypt/pebble), this leads to a JWS error, because 

- the real account URL is e.g. `https://localhost:14000/my-account/8`, whereas 
- the account URL submitted by dehydrated is `https://localhost:14000/sign-me-up/8`.

Rogdham:
Fixes #671. Note that I did not change the ACMEv1 code at all, not sure what that version of the protocol is supposed to do.

We store the account URL on account creation in the `account_id.json` file.

When reading the file, if the attribute is missing, we fall back to previous behaviour, with a warning message.

---

Per https://tools.ietf.org/html/rfc8555#section-7.3
> The server returns this account object in a 201 (Created) response, with the account URL in a Location header field.  The account URL is used as the "kid" value in the JWS authenticating subsequent requests by this account (see Section 6.2).  The account URL is also used for requests for management actions on this account, as described below.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -408,17 +408,17 @@ init_system() {
       ACCOUNT_URL="${CA_REG}/${ACCOUNT_ID}"
     else
       if [[ -e "${ACCOUNT_ID_JSON}" ]]; then
-        ACCOUNT_ID="$(cat "${ACCOUNT_ID_JSON}" | get_json_string_value id)"
-      else
-        echo "+ Fetching account ID..."
+        ACCOUNT_URL="$(cat "${ACCOUNT_ID_JSON}" | get_json_string_value url)"
+      fi
+      # if account URL is not storred, fetch it from the CA
+      if [[ -z "${ACCOUNT_URL:-}" ]]; then
+        echo "+ Fetching account URL..."
         ACCOUNT_URL="$(signed_request "${CA_NEW_ACCOUNT}" '{"onlyReturnExisting": true}' 4>&1 | grep -i ^Location: | awk '{print $2}' | tr -d '\r\n')"
-        ACCOUNT_ID="${ACCOUNT_URL##*/}"
-        if [[ -z "${ACCOUNT_ID}" ]]; then
+        if [[ -z "${ACCOUNT_URL}" ]]; then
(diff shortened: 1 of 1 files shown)


lukas2511:
Looks good, but it would be great if instead of showing a warning every time it would fallback silently and only throw an error if there is no account url when initially fetching account information,.

lukas2511:
Also on new accounts the ID wouldn't actually be required, and might even be against what the RFC describes, so it would be better to simply not store it.

So maybe a fallback logic like

if url exists: use url
elif id exists: use fixed accounts url with id
else fetch account url

Rogdham:
I believe I implemented all your suggestions.

However, I am not sure about the case `elif id exists: use fixed accounts url with id`: we could just fall through the `else fetch account url` case, and retrieve (and store) the account URL. This would mean transparently updating the `account_id.json` file from `{"id":"42"}` to `{"url":"https://something/blah/42"}`. Tell me if you would prefer that, it's only 6 lines to remove :wink:

lukas2511:
Yea, i guess you are right, it would probably be best to simply fetch the url and have everything up to standard. I was thinking about this way to complicated when I answered before...

Rogdham:
All right, I've updated my commit.

lukas2511:
Thanks again! And sorry for the slow response, I'm surprisingly busy setting up video infrastructure for my university at the moment...

Rogdham:
No worries, and thank you for merging all those PRs! :tada:

PR #735, opened 2020-05-14
Fix tls-alpn-01 configuration example

cyclopsian:
Using nginx 1.14, I get an error about `map` not being valid inside a `server` block. According to [the docs](https://nginx.org/en/docs/stream/ngx_stream_map_module.html#map) it's only valid in a `stream` block. I made the change on my server and it fixed it for me, this updates the example.

### DIFF
--- a/docs/tls-alpn.md
+++ b/docs/tls-alpn.md
@@ -15,18 +15,16 @@ Your config should look something like this:
 
 ```nginx
 stream {
-  server {
-    map $ssl_preread_alpn_protocols $tls_port {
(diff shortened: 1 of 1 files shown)


lukas2511:
You are right, I have no idea how that ended up like this. My config is a lot more complex than the basic example and I must have messed up copying the relevant parts... Thanks! Merged on current dev branch https://github.com/dehydrated-io/dehydrated/tree/jsonsh and will be merged to master soon(ish).

PR #748, opened 2020-07-20
Fix #747 by enabling keep-going by default, add --abort-on-error

### LINKED ISSUE #747: Enable `--keep-going` by default for cronjob-operations.
### What happened
We host a few wikis for random student's projects and allow them to use their own domain names.

Over the weekend some of our hosted wikis became unreachable (cert expired), because our cronjob was this:
`@weekly letsencrypt /usr/bin/dehydrated -c; /usr/bin/dehydrated -gc`

Turns out: dehydrated aborts by default on the first error and one domain expired and failed to be renewed, aborting the renew operation for all subsequent domains. (And we failed to check on the "error" mails from dehydrated/cron, because you get them even if no error occurs so it turns out: all of us admins ignore them)

### Proposal
Change the default of `--keep-going` to be enabled by default (and add a `--abort-on-error` or something like that to cover the usecase of actually wanting to fail on the first error.)

someone-somenet-org:
Set PARAM_KEEP_GOING to "yes" by default. Add `--abort-on-error` parameter.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -153,6 +153,7 @@ load_config() {
   DEHYDRATED_USER=
   DEHYDRATED_GROUP=
   API="auto"
+  PARAM_KEEP_GOING="yes"
 
   if [[ -z "${CONFIG:-}" ]]; then
     echo "#" >&2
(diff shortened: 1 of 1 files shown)


lukas2511:
Not going to make it the default, sorry. Thanks for the suggestion anyways.

Issue #665, opened 2019-08-20
setting DEHYDRATED_USER does not check uid first


jmgurney:
If you set DEHYDRATED_USER (to prevent accidentally running as root), it checks to make sure sudo is available before verifying the current user.  If you have a system w/o sudo, instead of running, it errors out:
```gold,pts,/usr/local/etc/dehydrated,5$/usr/local/bin/dehydrated -c
# INFO: Using main config file /usr/local/etc/dehydrated/config
ERROR: DEHYDRATED_USER set but sudo not available. Please install sudo.
gold,pts,/usr/local/etc/dehydrated,6$id
uid=443(_letsencrypt) gid=443(_letsencrypt) groups=443(_letsencrypt)
gold,pts,/usr/local/etc/dehydrated,7$grep DEHYDRATED_USER /usr/local/etc/dehydrated/config
DEHYDRATED_USER=_letsencrypt
```
expected behavior: That when the script is running as the specified DEHYDRATED_USER (and _GROUP), and sudo is not installed, that the script will run as expected.

PR #758, opened 2020-09-01
sudo: fix for #665 and avoid silent failure on error

danimo:


txr13:
In has_sudo, you’re checking for “sud”, not “sudo”.

danimo:
@txr13 Indeed. Remainer of a test. Fixed.

PR #760, opened 2020-09-01
implement account deactivation through --deactivate parameter

### LINKED ISSUE #216: Feature: Deleting an Account
ACME allows deletion of accounts, this may be implemented in letsencrypt.sh in the future

danimo:
This is an updated version of https://github.com/lukas2511/dehydrated/files/2641548/dehydrated_add_deactivate_command.diff.txt

Fixes #216

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,10 @@
 This file contains a log of major changes in dehydrated
 
 ## [x.x.x] - xxxx-xx-xx
+## Added
+- Support for external account bindings
+- Special support for ZeroSSL
+
 ## Fixed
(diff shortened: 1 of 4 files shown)


lukas2511:
This should do something with the deactivated key so it doesn't just error out on rerun. Maybe touch a file "deactivated" in the account directory and check on load if it exists, if it does error out "Account has been deactivated, please delete old key and register a new one.", or just simply delete or rename the account directory after it was successfully deactivated.

Also as stated in #216 ACMEv1 does support deactivation as well and it should be easy to implement as it's basically the same as with v2.

lukas2511:
Merged

PR #761, opened 2020-09-07
Replace \s with [[:space:]] for compatibility

oranenj:
Replacing `\s` with `[[:space:]]` in the various `sed` and `grep` invocations makes dehydrated work on OpenBSD.

This is *very* lightly tested though; I only tested the `dns-01` challenge on OpenBSD and Alpine Linux. However, to my knowledge GNU utils handle these just fine.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -45,27 +45,27 @@ json_path() {
 # Get string value from json dictionary
 get_json_string_value() {
   local filter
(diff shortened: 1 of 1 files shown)


jobe1986:
I can confirm this works on FreeBSD 12.1, and doesn't appear to cause issue on my Ubuntu boxes.

lukas2511:
Merged. Thx!

PR #788, opened 2020-12-06
force-renew certificates when switching ACME providers

pyke369:
useful when switching CA for an existing certificates tree

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1645,6 +1645,23 @@ command_sign_domains() {
       fi
     fi
 
+    # Force-renew if certificate was signed by a different provider
+    provider="${certdir}/provider"
+    if [[ ! -s "${provider}" ]]; then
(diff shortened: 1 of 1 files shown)


lukas2511:
I don't think this is useful. Just because you're switching CAs doesn't necessary make the thousands of already issued certificates invalid. If you really want this you can always manually force the renewal process.

pyke369:
This is actually because we maintain different domains.txt referenced by different configuration files with different CAs on top of a unique certificates store tree, so we can choose which CA a particular certificate is managed by (we do that through a SaltStack automation and we don't renew associated private keys). Mainly because the CA variable cannot be overloaded by a DOMAINS_D or certificate folder extra configuration file at this time.

Also forcing the renewal process will force it for all domains in a given domains.txt, even if only a few of them were switched to another CA.

PR #798, opened 2021-02-02
Per-certificate config fixes

lespinasse:
- Ensure that all per-certificate settings are saved and restored in
  store_configvars() and reset_configvars() - that's what makes them
  per-certificate in the first place...

- Add OCSP_FETCH and OCSP_DAYS in the documented list of supported
  per-certificate configs, since the code does allow these.


lukas2511:
Thx, merged :)

PR #804, opened 2021-02-22
document using -t tls-alpn-01 with lighttpd

gstrauss:
document using `-t tls-alpn-01` with lighttpd

lighttpd supports TLS-ALPN-01 without the need for a separate `acme-tls/1` responder.


lukas2511:
Cool, thx.

PR #810, opened 2021-03-19
Support reading domains from drop-in snippets in `domains.txt.d`

schuetzm:
This adds support for a directory `domains.txt.d` in the same location as `domains.txt`, which can contain further files in the same format as `domains.txt` that are automatically read and appended to the list of domains.

This follows the common pattern of config drop-in directories that is used by e.g. Nginx, Apache, systemd etc. It allows adding new domains by simply dropping a new file in the directory instead of editing an existing file, which is useful for automation and scripting.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1532,6 +1532,20 @@ command_account() {
   exit 0
 }
 
+# Parse contents of domains.txt and domains.txt.d
+parse_domains_txt() {
+  # Allow globbing temporarily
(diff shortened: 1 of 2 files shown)


jobe1986:
Honestly I like this idea in terms of usefulness to automated systems.

lukas2511:
I've decided to merge this, but added a note that behaviour of this might change. The naming is a bit confusing as there already is a `DOMAINS_D` config variable which is used for per-certificate configurations. Now there's basically `domains.txt.d` and `domains.d` :see_no_evil:

PR #815, opened 2021-04-12
ensure newline before new section in openssl.cnf

27o:
openssl.cnf may not end with a newline. The section [SAN] will then not be found as it is added to the last line of openssl.cnf.


lukas2511:
Thx, merged.

PR #840, opened 2021-09-20
Better handling around grep/awk

### LINKED ISSUE #389: --cleanup from hook
is it possible to call `dehydrated --cleanup` from within a hook, e.g. `exit_hook`?

### LINKED ISSUE #839: register fails when egrep doesn't support -a
The detection for which grep to use doesn't take into account that -a is not always supported, which leads to it not setting `GREP=awk_egrep`.

```
# ./dehydrated --register --accept-terms
# INFO: Using main config file /opt/dehydrated/config
egrep: illegal option -- a
usage: egrep [-bchHilLnoqrRsvx] [-A num] [-B num] [-C num|-num]
             [--label=name] [-e pattern_list]... [-f pattern_file]...
             [pattern_list] [file]...
EXPECTED value GOT EOF
```

Additionally, systems where egrep doesn't support `-a`, probably won't have `gawk`. But even if they do, gawk produces this error:

```
gawk: cmd. line:3: (FILENAME=- FNR=1) warning: regexp escape sequence `\u' is not a known regexp operator
```

or

```
gawk: cmd. line:3: (FILENAME=- FNR=1) fatal: Invalid range end: /"[^[:cntrl:]"\]*((\[^u[:cntrl:]]|\u[0-9a-fA-F]{4})[^[:cntrl:]"\]*)*"|-?(0|[1-9][0-9]*)([.][0-9]*)?([eE][+-]?[0-9]*)?|null|false|true|[[:space:]]+|./
```

(depending on version).

bahamat:
This fixes #839.

* Test `egrep -ao` instead of just `-o` to avoid selecting `egrep` when `-a` is not supported  (i.e., gnu-grep egrep is not available)
* Use `awk` instead of `gawk`. `gawk` doesn't support the escape sequence passed, so it wouldn't anyway. But Non-GNU systems will usually have a different flavor of awk where the escape sequence *does* work.

This has been tested on:

* CentOS 7
* CentOS 8
* Debian 10
* FreeBSD
* illumos (SmartOS)
* Ubuntu 20.04

I can also test on others if that's helpful.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -88,7 +88,7 @@ jsonsh() {
   awk_egrep () {
     local pattern_string=$1
 
-    gawk '{
+    awk '{
       while ($0) {
         start=match($0, pattern);
         token=substr($0, start, RLENGTH);
@@ -110,7 +110,7 @@ jsonsh() {
       GREP='egrep -ao'
     fi
 
-    if echo "test string" | egrep -o "test" >/dev/null 2>&1
+    if echo "test string" | egrep -ao "test" >/dev/null 2>&1
     then
       ESCAPE='(\\[^u[:cntrl:]]|\\u[0-9a-fA-F]{4})'
       CHAR='[^[:cntrl:]"\\]'


dehydrated
```diff
@@ -17,7 +17,7 @@ umask 077 # paranoid umask, we're creating private keys
 exec 3>&-
 exec 4>&-
 
-VERSION="0.7.1"
+VERSION="0.7.2"
```
glensc, line 20: version change is maintainer only, not from pull requests

bahamat: Reverted that line.

lukas2511:
thx, merged

PR #846, opened 2021-10-17
Fix command_version on Darwin/macOS 11

jcmoraisjr:
Current output of `uname` on Darwin/macOS 11 is only `Darwin`, which breaks the premisse used in `command_version()`. This update adds `Darwin` alongside `BSD`.


lukas2511:
thx, merged

PR #847, opened 2021-10-17
Fix hook based challenges if the same domain appears twice or more

jcmoraisjr:
When two or more domains are present in the challenge, all the hook calls happen before asking the acme server to verify the challenge. This is breaking configurations that need to verify the same domain more than once, e.g. dns based challenges with `example.com`. and `*.example.com`. In this case, `example.com` will be verified twice, maybe with distinct tokens, leading to the second configuration overwriting the first one before the challenge verification.

This update moves the hook call into the while-loop that asks the acme server to verify the challenge, so a second configuration will only happen after the first one has being verified.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1080,23 +1080,19 @@ sign_csr() {
   echo " + ${num_pending_challenges} pending challenge(s)"
 
   # Deploy challenge tokens
-  if [[ ${num_pending_challenges} -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


txr13:
https://github.com/dehydrated-io/dehydrated/blob/master/docs/troubleshooting.md#dns-invalid-challenge-since-dehydrated-060--why-are-dns-challenges-deployed-first-and-verified-later

The current behavior is by design. Having multiple TXT records for the same domain name (eg. two different instances of `_acme-challenge.example.com`) is supported by most DNS hosting providers who adhere to spec.

PR #848, opened 2021-10-27
Make CHALLENGETYPE available in Hook.

basti12203:


lukas2511:
`CHALLENGETYPE` is exported early on in the script and is already available as environment variable in the hook script. Is there any special reason why you need it as direct argument to the hook script? I'd like to avoid altering hook parameters where possible.

basti12203:
Thx for the Info. So ignore my pull request.

PR #868, opened 2022-03-23
egrep is deprecated

simondeziel:
egrep has been deprecated since 2007 and warns it's obsolete since:
https://git.savannah.gnu.org/cgit/grep.git/commit/?id=a9515624709865d480e3142fd959bccd1c9372d1

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -125,8 +125,7 @@ jsonsh() {
       GREP='grep -Eao'
     fi
 
-    # shellcheck disable=SC2196
-    if echo "test string" | egrep -ao "test" >/dev/null 2>&1
+    if echo "test string" | grep -Eao "test" >/dev/null 2>&1
     then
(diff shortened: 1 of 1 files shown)


dehydrated
```diff
@@ -126,7 +126,7 @@ jsonsh() {
     fi
 
     # shellcheck disable=SC2196
```
txt-file, line 128: The shellcheck comment/command should be removed as it is not needed anymore.

txt-file (changes requested):

dehydrated
simondeziel: Thanks, done.

txt-file: That was fast. :smile:

danimo (approved):

lukas2511:
merged to master branch, thx

PR #877, opened 2022-06-07
Add check for head/tail

krayon:
head/tail are used but not checked for


krayon:
This is encompassed in #878

PR #890, opened 2022-10-20
egrep -> grep -E

### LINKED ISSUE #889: egrep: warning: egrep is obsolescent; using grep -E
Thanks for this great tool, have been using it for a while, now.

Recently, I’ve started seeing warnings that `egrep` should be replaced with `grep -E`. I’m using **grep-3.8**.

For now, this is of course nothing more than a nuissance, but I would think that eventually, distributions will stop shipping an `egrep`.

christophfink:
Prevent the warnings described in issue #889. I only found one occurrance of `egrep` being used.

(closes #889 )

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -126,7 +126,7 @@ jsonsh() {
     fi
 
     # shellcheck disable=SC2196
-    if echo "test string" | egrep -ao "test" >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


lukas2511:
fix was already in testing branch, merged together with a few other changes to master now, this issue should be fixed

PR #915, opened 2023-06-14
Proxy-option for openssl-ocsp stapling

NiceRath:
Greetings!

This PR adds the missing proxy functionality for the openssl 'ocsp_log' call.
See issue: https://github.com/dehydrated-io/dehydrated/issues/838

What does it do?
It checks if a proxy is set in the environment variables and uses it for the ocsp-call if so.

\- Rath

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -289,6 +289,7 @@ store_configvars() {
   __WELLKNOWN="${WELLKNOWN}"
   __HOOK_CHAIN="${HOOK_CHAIN}"
   __OPENSSL_CNF="${OPENSSL_CNF}"
+  __OPENSSL_PROXY="${OPENSSL_PROXY}"
   __RENEW_DAYS="${RENEW_DAYS}"
   __IP_VERSION="${IP_VERSION}"
 }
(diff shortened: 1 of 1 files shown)


bllfr0g:
wouldn't it be easier/cleaner to use the `-proxy` option to openssl-ocsp?

NiceRath:
Greetings!

I thought so too - but It seems not all OpenSSL packages have this functionality implemented.
That's probably why the problem occurs in the first place..

Per example when using OpenSSL '1.1.1n-0+deb11u4' one gets the following error:
`ERROR: Error while fetching OCSP information: ocsp: Unrecognized flag proxy`

\- Rath

bllfr0g:
good point! Looks like `-proxy` was added about two years ago, so your change will work in more places.

﻿﻿﻿﻿﻿﻿﻿﻿https://github.com/openssl/openssl/commit/88d96983d881254d0bcb36d79b32aac08339e0d3

NiceRath:
It seems there is an issue with my solution.
'set -u' currently stops the scripts execution if the environmental proxy-variables are unset.
I'll test it and add a fix to the PR.

\- Rath

PR #923, opened 2023-12-03
Ignore output of 'openssl req -verify'.

wteiken:
Newer versions of openssl seem to send the verify outout to stdout instead of stderr in the past. Ignore that output when retrieving altnames.


lukas2511:
merged in 4fd777e87e589652b1127b79ac6688ed7cb151fe, thx

PR #929, opened 2023-12-17
Suppress openssl warning about reading from stdin.

wteiken:
Fixed some more locations where warnings about reading from stdin showed up.

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1015,7 +1015,7 @@ extract_altnames() {
     _exiterr "Certificate signing request isn't valid"
   fi
 
-  reqtext="$( <<<"${csr}" "${OPENSSL}" req -noout -text )"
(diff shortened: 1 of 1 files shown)


Gnarfoz:
Instead of redirecting stderr, wouldn't supplying an explicit `-in /dev/stdin` be cleaner?

wteiken:
> Instead of redirecting stderr, wouldn't supplying an explicit `-in /dev/stdin` be cleaner?

Not sure it is as compatible, but would probably also work.

It seems for openssl they are acknowledging that the error message was a bad idea will only output if stdin is actually a tty, so this will hopefully become obsolete.

wteiken:
Closing in favor of #940 now that "-in -" is supported.

PR #952, opened 2024-12-29
Correct indentation of visible output

robert-scheck:
Attempt to upstream https://src.fedoraproject.org/rpms/dehydrated/blob/rawhide/f/dehydrated-autowash.patch


lukas2511:
Added as part of bigger commit cf9e6a33fd3ed0576e2f0d14f20c4a96909c1fee, thanks.

PR #954, opened 2024-12-29
Include local overrides for `hook.sh` functions

robert-scheck:
Attempt to upstream https://src.fedoraproject.org/rpms/dehydrated/blob/rawhide/f/dehydrated-hook.sh-defaults.patch partially

### DIFF
--- a/docs/examples/hook.sh
+++ b/docs/examples/hook.sh
@@ -214,6 +214,13 @@ exit_hook() {
   #   Contains error message if dehydrated exits with error
 }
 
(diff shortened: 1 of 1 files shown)


lukas2511:
This makes a few too many assumptions for my liking (fixed configuration path, assumes script files being parsable by the same shell which launched the hook script in the first place). Even though it's just an example I'd like to keep it a bit more generic. I have a different version in mind, but will keep this PR open for tracking purposes. And sorry for the late response.

PR #956, opened 2025-01-14
implemented certificate profile selection (draft-aaron-acme-profiles-00)

zhangyoufu:
Let's Encrypt announced Certificate Profile Selection feature. A profile describes a collection of attributes about the certificate that will be issued, such as what extensions it will contain, how long it will be valid for, and more.

This feature is currently available in Let's Encrypt staging environment. ACME clients are encouraged to introduce support for this new field.

https://letsencrypt.org/2025/01/09/acme-profiles/
https://datatracker.ietf.org/doc/html/draft-aaron-acme-profiles-00

---

In this PR, this feature is implemented as global configuration, per-certificate configuration and command line parameter.

In case a profile is not specified, the behavior remains unchanged.

When a profile is specified but not available from the CA:
```
ERROR: ACME profile 'test' not found, available profiles:
  classic: The same profile you're accustomed to
  tlsserver: https://letsencrypt.org/2025/01/09/acme-profiles/
```

When a profile is specified but CA does not provide any valid profiles:
```
ERROR: ACME profile not supported by this CA
```

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -291,6 +291,7 @@ store_configvars() {
   __OPENSSL_CNF="${OPENSSL_CNF}"
   __RENEW_DAYS="${RENEW_DAYS}"
   __IP_VERSION="${IP_VERSION}"
+  __ACME_PROFILE="${ACME_PROFILE}"
 }
 
 reset_configvars() {
@@ -309,6 +310,7 @@ reset_configvars() {
   OPENSSL_CNF="${__OPENSSL_CNF}"
   RENEW_DAYS="${__RENEW_DAYS}"
   IP_VERSION="${__IP_VERSION}"
+  ACME_PROFILE="${__ACME_PROFILE}"
 }
 
 hookscript_bricker_hook() {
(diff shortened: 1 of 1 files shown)


Weishaupt:
@lukas2511 would love to see this integrated into dehydrated. I'm eager to try out short lived certs with the staging environment.

lukas2511:
> @lukas2511 would love to see this integrated into dehydrated. I'm eager to try out short lived certs with the staging environment.

I’d really like to see this as well! I’m currently wrapping up my master’s thesis, but once that’s submitted, I’ll review this PR and some others.

lukas2511:
I've merged this (5ab8c3806d78f8ff92439a7e305b7288c3b4ad21) and added changelog (1a1cb94a614bc3944c83e346c400c8a60311720e) as well as a small fix to the logic which was checking for profile validity (f6d82e271589a2d9b2edf98dfc4dab04b35d7919). Thanks :)

PR #966, opened 2025-05-12
check for existence of ocsp responder if OCSP_FETCH="yes" is set

bjacke:
this allows having globally OCSP_FETCH="yes" and not failing for certificates that don't have any ocsp support at all (like latest letsencrypt certs...)

### DIFF
--- a/dehydrated
+++ b/dehydrated
@@ -1945,7 +1945,7 @@ command_sign_domains() {
         update_ocsp="yes"
       fi
 
-      if [[ "${update_ocsp}" = "yes" ]]; then
+      if [[ "${update_ocsp}" = "yes" && -n $("${OPENSSL}" x509 -in "${cert}" -noout -ocsp_uri) ]]; then
         echo " + Updating OCSP stapling file"
         ocsp_timestamp="$(date +%s)"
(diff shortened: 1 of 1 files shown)


lukas2511:
Mh I think I'd rather have it fail, but a bit more graceful like it seems to fail at the moment (even breaking the "keep going" behaviour). This is definitively an error case one might want to catch, especially if ocsp stapling was previously configured in e.g. a webserver.

bjacke:
I used to use ocsp with must staple on almost all systems. What I can tell is that none of the server software fails to start up with a non-matching ocsp response file. It might give a warning but it does not fail to start.
With letsencrypt stopping to use ocsp there there is a hard fail now with dehydrated because new certificates just don't get successfully requested any more.
A good middle way might be that dehydrated also issues a warning on STDERR, so that cron jobs or so generate a warning.
Just failing to issue new certs is the worst option imho because work flows that used to work since aged suddenly fail hard. And then it takes time until the resposible person is available, find the problem and fixes it. That would really be a sub-optimal behaviour.

lukas2511:
I've decided to go a slightly different route:
- Throw error with information about deprecation instead of letting openssl run into an error
- Also skip over ocsp stapling errors when keep-going is set

Simply skipping over the error without any manual intervention didn't seem like a good idea to me.

PR #972, opened 2025-06-01
Don't allow CDN's to send cached responses

GTAXL:
A lot of CA's use a CDN service to protect and speed up their ACME service. These CDN services can sometimes miss-behave and send cached results. For example DigiCert's ACME service uses the Imperva CDN. It will send cached results on the DNS validation, challenge endpoint, resulting in it being stuck in the processing status, thus dehydrated is hung and never gets the certificate.


lukas2511:
Not sure if this really makes a change, but shouldn't hurt. Merged manually.

PR #995, opened 2026-03-31
Update zeroapi curl call to use ip_version + CURL_OPTS + dehydrated user-agent

radub2012:
ZeroSSL API curl request is not using the ip_version, CURL_OPTS, and user-agent for EAB requests, as the other curl's in the script. In evironments where ip_version or CURL_OPTS are needed to connect to zeroapi's SSL, the lack of these will trigger a failure in registering new accounts with ZeroSSL's EAB


lukas2511:
thanks, i've merged those changes :)

PR #996, opened 2026-03-31
Fix SAN comparison

domrim:
During testing with certificates containing DNS, IPv6 and IPv4 adresses we found some bugs in the comparison logic for existing certificates. Certificates were always regenerated as the normalized output of the openssl command and the domains.txt-line did not match. Also we found a small typo in the grep-regex.


lukas2511:
thanks, i've merged your changes

